PR #9, opened 2020-02-06
Feature/arch 105

tinydylan:
See also https://github.com/tinymce/waluigi/pulls

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 #!groovy
-@Library('waluigi@feature/ARCH-105') _
+@Library('waluigi') _
 
 // NOTE: This Jenkinsfile relies on Tiny's internal infrastructure
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,30 +16,27 @@ node("primary") {
     checkout localBranch(scm)
   }
 
-  withBitbucket {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -24,30 +16,27 @@ node("primary") {
...
+  }
 
-    if (isReleaseBranch()) {
-      stage("publish") {
-        sshagent(credentials: ['8aa93893-84cc-45fc-a029-a42f21197bb3']) {
-          sh 'yarn lerna publish from-package --yes --ignore @ephox/bedrock-sample'
-        }
+  if (isReleaseBranch()) {
+    stage("publish") {
+      sshagent(credentials: ['8aa93893-84cc-45fc-a029-a42f21197bb3']) {
```
spocke: Is it possible to have these credentials in a constant in waluig? It seems to be used all over the place since it's a guid would that be different if the jenkins server would need to be re-installed?

tinydylan: Actually... that GUID is the stash credentials. This would be used so lerna can commit back to stash.  The build server would already have these credentials in an ssh agent. We can just delete the sshagent call here.

I'll look at the other credentials entries as I come to them. If we need a constant in waluigi, I'll add it.

PR #13, opened 2020-02-17
TINY-4696: Fixed test files being cached incorrectly and added new `wipeBrowserCache` flag

lnewson:
I found a couple of endpoints that didn't set appropriate cache headers, which I believe is what was causing IE 11 to cache the TinyMCE tests. This will now set both an `ETag` and the `Cache-Control: public, max-age=0` response headers which will should make it revalidate on each load.

Note: I opted to make the IE cache wipe opt in, as it's a destructive behaviour so it's better to not have something destructive as a default option.


modules/server/src/main/ts/bedrock/server/Routes.ts
```diff
@@ -15,6 +16,20 @@ const createServer = (root: string) => {
...
 
+const doResponse = (request: IncomingMessage, response: ServerResponse, status: number, contentType: string, data: any) => {
+  response.setHeader('ETag', RouteUtils.generateETag(data));
+  if (status === 304 || RouteUtils.isCachable(status) && RouteUtils.isFresh(request, response)) {
+    response.writeHead(304, { });
+    response.end();
+  } else {
+    response.writeHead(status, {
+      'Content-Type': contentType,
+      'Cache-Control': 'public, max-age=0'
```
tiny-james: Why did we choose max-age over no-cache?

tiny-james: Answering myself:
https://stackoverflow.com/a/1383359/7377237
Seems that max-age allows the browser to use a stale copy where as no-cache requires the browser to always fetch a new copy which might be a little too aggressive.

lnewson: To be completely honest I did this, as that's what `serve-static` was using, so I just wanted to keep it consistent.

PR #14, opened 2020-02-20
TINY-4713: Ensure tests aren't skipped due to parsing errors and add the symbol polyfill

lnewson:
We found out today that a number of tests weren't running on IE 11 (65 out of 782). This was due to `fast-check` being added as a dependency and it makes use of `Symbol` which is ES6 and as such was throwing an uncaught exception when importing the related tests on IE 11.

So this makes it so each test is wrapped in a try/catch and will now log a test failure instead. It also adds the `Symbol` polyfill from `core-js` to make it possible to run the tests on IE 11.


modules/server/src/main/ts/bedrock/compiler/Imports.ts
```diff
@@ -15,17 +15,28 @@ const filePathToImport = (useRequire: boolean, scratchFile: string) => {
     filePath = filePath.replace(/\\/g, '/');
 
     return [
-      useRequire ? `require("${relativePath}");` : `import "${relativePath}";`, // rollup doesn't support require
-
-      `addTest("${filePath}");`
+      'try {',
```
ephox-mogran: How interesting. Is "```" an option?

modules/server/src/main/ts/bedrock/compiler/Imports.ts
```diff
@@ -52,16 +63,20 @@ const addTest = (testFilePath: string) => {
     console.error('no test list to add tests to');
   }
 };
+const handleParseError = (testFilePath: string, error: any) => {
+  ${useRequire ? 'const UnitTest = require(\'@ephox/bedrock-client\').UnitTest;' : 'import { UnitTest } from \'@ephox/bedrock-client\';'}
```
ephox-mogran: Given the amount of escaping failures possible in statements like these, could we load these snippets from a file? Could that make it easier?

ephox-mogran: I guess the parameterisation of it probably makes it harder.

modules/server/src/main/ts/bedrock/compiler/Imports.ts
```diff
@@ -85,10 +100,14 @@ var addTest = function (testFilePath) {
     console.error('no test list to add tests to');
   }
 };
+var handleParseError = function (testFilePath, error) {
```
ephox-mogran: This looks almost identical to the above. Can the string be shared?

lnewson: It's not identical at all, one is `JS` and the other is strict `TS`, so no they can't be shared sorry.

modules/server/src/main/ts/bedrock/compiler/Imports.ts
lnewson: Yeah, most of this is dynamic and hence can't be in a file. We could look at making a template file for this, but I'm not convinced it's worth it as it'd still run into similar issues.

modules/server/src/main/ts/bedrock/compiler/Imports.ts
lnewson: More than likely, this was probably from before ES 6 was possible on node and has just been adjusted slowly over time.

modules/server/src/main/ts/bedrock/compiler/Imports.ts
ephox-mogran: Ah. I see.

ephox-mogran: What benefit do you get from making the TS one TS? Can it pick up any errors at that point? Would you be able to share it if they were both JS? Probably a stupid question, I'm just genuinely curious.

ephox-mogran: Or is the TS one required because there is a strict setting on?

lnewson: The TS one is required as bedrock reads from the projects tsconfig, so if strict is on it'll complain and fail to compile.

modules/server/src/main/ts/bedrock/compiler/Imports.ts
```diff
@@ -52,16 +63,20 @@ const addTest = (testFilePath: string) => {
...
+const handleParseError = (testFilePath: string, error: any) => {
+  ${useRequire ? 'const UnitTest = require(\'@ephox/bedrock-client\').UnitTest;' : 'import { UnitTest } from \'@ephox/bedrock-client\';'}
+  UnitTest.test(testFilePath, () => { throw error; });
+};
 `,
     imports,
-    'export {};'
-  ].join('\n');
+    '\nexport {};'
+  ]).join('\n');
```
tinydylan: Would multiline interpolated strings help here?

PR #15, opened 2020-02-26
fixed import generation for js tests

spocke:
I noticed that the import file was corrupt when updating bedrock to the latest version.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,7 @@
+Version 9.5.1
+
+* Fixed broken import generation for js tests.
+
 Version 9.5.0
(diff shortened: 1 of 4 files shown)


lnewson:
Thanks heaps for fixing this @spocke I knew something was broken since it was getting caught in an infinite compilation loop, but others things have prevented me from looking into it (ie pricing refresh). 👍

lnewson:
Although in saying that, my issue was with TS in compilation TinyMCE so that's probably a separate issue after looking at the actual changes 🤷‍♀️

PR #16, opened 2020-02-27
TINY-4713: Fixed bedrock ending up in an infinite compilation loop

lnewson:
I'm not entirely sure why, but it seems as though webpack or one of the loaders we're using ends up in an infinite compilation loop when there's too many try/catch statements.

Previously we had a try/catch statement for each test, however to prevent the infinite loop I've changed this back to just one global catch all. This does mean other tests can't run if one has an import error, but at least with this change we'll still know that tests failed to run.

Here's an example of what the failure looks like:
```
-------------------------------------------------------------------------------
Test failed: Import Error - src/test/ts/client/SyncPassTest.ts (undefined)
TypeError: undefined is not a function (evaluating 'window.blah()')

Logs:


-------------------------------------------------------------------------------
Some tests failed. See {scratch/TEST-bedrock-run.xml} for details.
```

### DIFF
--- a/modules/server/src/main/ts/bedrock/compiler/Imports.ts
+++ b/modules/server/src/main/ts/bedrock/compiler/Imports.ts
@@ -62,7 +62,8 @@ const addTest = (testFilePath: string) => {
 };
 const handleImportError = (testFilePath: string, error: any) => {
   ${useRequire ? 'const UnitTest = require(\'@ephox/bedrock-client\').UnitTest;' : 'import { UnitTest } from \'@ephox/bedrock-client\';'}
-  UnitTest.test('Import Error - ' + testFilePath, () => { throw error; });
+  UnitTest.test('Import Error', () => { throw error; });
+  addTest(testFilePath);
 };
 
 let __currentTestFile;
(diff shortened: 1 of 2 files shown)


spocke:
I wonder if we could use https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror

lnewson:
> I wonder if we could use https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror

Good idea, I'd forgotten about that. Looks to work fine, so I've updated to that :+1:

modules/server/src/main/ts/bedrock/compiler/Imports.ts
```diff
@@ -36,12 +33,13 @@ const generatePolyfills = (useRequire: boolean): string[] => {
 const generateImportsTs = (useRequire: boolean, scratchFile: string, srcFiles: string[]) => {
   const imports = srcFiles.map(filePathToImport(useRequire, scratchFile)).join('\n');
   // header code for tests.ts
-  return generatePolyfills(useRequire).concat([
-    `
+  return `${generatePolyfills(useRequire)}
```
jhaines: is this missing the closing `?

PR #18, opened 2020-03-17
Upgrade dependencies to fix security alerts

lnewson:


modules/server/src/test/ts/AttemptUtils.ts
```diff
@@ -9,7 +9,7 @@ export const arbAttemptPassed = <E, A> (arba: Arbitrary<A>): Arbitrary<Attempt<E
   arba.map<Attempt<E, A>>(Attempt.passed);
 
 export const arbAttempt = <E, A> (arbe: Arbitrary<E>, arba: Arbitrary<A>): Arbitrary<Attempt<E, A>> =>
-  fc.oneof(arbAttemptFailed(arbe), arbAttemptPassed(arba));
+  fc.oneof(arbAttemptFailed<E, A>(arbe), arbAttemptPassed<E, A>(arba));
```
tinydylan: See, this right here is why people like [Hindley-Milner-Damas type systems](https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system)

PR #19, opened 2020-03-30
Restored JS source maps

TheSpyder:
Follow up to TINY-4269, which only fixed TypeScript source maps (and the "unused param" was later removed)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @spocke @jhaines @metricjs @lnewson @ltrouton @tinydylan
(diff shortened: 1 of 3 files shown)


lnewson:
Also thanks for fixing this, I should have noticed it when I fixed the TS issue! 👍

CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @spocke @jhaines @metricjs @lnewson @ltrouton @tinydylan
```
lnewson: Did you want to include yourself in this list?

TheSpyder: not for the moment, no :)

PR #23, opened 2020-06-19
Bedrock improvements

tinydylan:
Related case: TINY-6136

Changes:
* Added `promiseTest` to test async functions (functions returning promises)
* Changed `asynctest` to `asyncTest`, following camelCase conventions. Old spelling remains for compatibility.
* Changed `asyncTest` so that it fails (instead of timing out) when an exception is thrown during execution of the test method.
* Remove the sample "suite" tests. The failing ones were causing timeouts.

Will release as new minor version: 9.6.0


modules/client/src/main/ts/api/UnitTest.ts
```diff
@@ -115,3 +125,9 @@ export const test = (name: string, test: () => void) => {
     }
   });
 };
+
+/** Tests an async function (function that returns a Promise). */
+export const promiseTest = (name: string, test: () => Promise<void>) =>
+  asyncTest(name, (success, failure) => {
+    test().then(success).catch(failure);
```
TheSpyder, line 132: We've had this in the past, I'm not sure why it was removed. Polyfill required on IE11, anyway.

lnewson:
Removed my changes requested, as per the discussion on slack. Not approving though, because of the issue Andy mentioned.

modules/client/src/main/ts/api/UnitTest.ts
tinydylan: As discussed out-of-band, we'll leave it up to the caller to polyfill Promises.

modules/sample/src/test/ts/client/AsyncFailTest.ts
```diff
@@ -1,10 +1,26 @@
 import { UnitTest } from '@ephox/bedrock-client'
 
-UnitTest.asynctest('AsyncFail Test', (success, failure) => {
+UnitTest.asyncTest('AsyncFail Test 1', (success, failure) => {
```
lnewson, line 3: This test still needs to import a promise polyfill from somewhere.

- [ ] Add Promise polyfill in this test

modules/sample/src/test/ts/client/AsyncPassTest.ts
```diff
@@ -1,10 +1,24 @@
 import { UnitTest } from '@ephox/bedrock-client'
 
-UnitTest.asynctest('AsyncPass Test', (success, failure) => {
-
+UnitTest.asyncTest('AsyncPass Test 1', (success, failure) => {
   new Promise(function (resolve, reject) {
```
lnewson, line 4: Interesting that this didn't have a poyfill previously... it really should have.

tinydylan: Yeah, looks like CI is only running these tests in phantomjs.

tinydylan: I've logged internal ticket TINY-6159 to add tests cross-browser.

tinydylan: When I add `@ephox/wrap-promise-polyfill` and try to import it, I get:

```
TS2529: Duplicate identifier 'Promise'. Compiler reserves name 'Promise' in top level scope of a module containing async functions.
```

tinydylan: I'm going to add the polyfill fix to TINY-6159. Sound good?

modules/client/src/main/ts/api/UnitTest.ts
TheSpyder: That's fine for tests _using_ bedrock, but not bedrock's tests.

modules/sample/src/test/ts/client/AsyncPassTest.ts
TheSpyder: That's a typescript error? weird, we don't have that issue in other tests, oh well.

tinydylan: As discussed: polyfill has been added. It doesn't work with `async..await` - seems like TS requires a global `Promise` for that. For now, I've removed all `async..await`.

PR #24, opened 2020-06-22
TINY-6147: Fixed incorrect test times being reported

lnewson:
There are two commits here: the first is just the simple fix for this and then the second is a restructure to try and reduce the monolith that is the reporter.

I was originally trying to restructure so I could write some tests, but since it relies on the DOM that turned out to be more trouble than it was worth.


modules/runner/src/test/ts/UtilsTest.ts
```diff
@@ -0,0 +1,22 @@
+import { assert } from 'chai';
```
ephox-mogran, line 1: Nice.

modules/runner/src/main/ts/core/Utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  const baseUrl = window.location.protocol + '//' + window.location.host + window.location.pathname;
+  if (offset > 0) {
+    const rt = (retry > 0 ? '&retry=' + retry : '');
+    return baseUrl + '?session=' + session + '&offset=' + offset + '&failed=' + failed + rt;
+  } else {
+    return baseUrl;
+  }
+};
+
+export const formatElapsedTime = (start: Date, end: Date): string => {
```
ephox-mogran, line 13: This code reminded me of `leftpad`, and the whole problem it caused. So I was curious to see what its current state was: https://www.npmjs.com/package/left-pad

It claims it's been superseded by https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart 

Do we trust that function? It looks like it isn't supported in IE? With the new separation of the parts of bedrock, I'm assuming `runner` is something that would run on the actual browser so it would need to work for IE?

modules/runner/src/main/ts/core/Utils.ts
```diff
@@ -0,0 +1,22 @@
+export const noop = (): void => {};
+
+export const makeUrl = (session: string, offset: number, failed: number, retry: number): string => {
+  const baseUrl = window.location.protocol + '//' + window.location.host + window.location.pathname;
+  if (offset > 0) {
+    const rt = (retry > 0 ? '&retry=' + retry : '');
+    return baseUrl + '?session=' + session + '&offset=' + offset + '&failed=' + failed + rt;
```
ephox-mogran, line 7: Minor: `baseUrl` is not required for calculating the suffix / query parameters. So the function could just be `getQueryParams` which would return '' or this query parameter-based suffix above. Then the calculation of `baseUrl` could be done elsewhere. Not sure if you'd want that, though. Makes it a bit easier to test, because you wouldn't have window (or an equivalent object passed in).

modules/runner/src/main/ts/core/Reporter.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  const summary = () => ({
+    passed: passCount + (params.offset - params.failed),
+    failed: failCount + params.failed,
+  });
+
+  const current = $('<span />').addClass('progress').text(params.offset);
+  const restartBtn = $('<button />').text('Restart').on('click', () => {
+    const url = makeUrl('', 0, 0, 0);
+    window.location.assign(url);
```
ephox-mogran, line 38: Minor: you could separate out the DOM entirely and pass in the handlers for the commands, and have setters for the labels.

modules/runner/src/main/ts/core/Reporter.ts
```diff
@@ -0,0 +1,163 @@
...
+  });
+  const retryBtn = $('<button />').text('Retry').on('click', () => {
```
ephox-mogran, line 40: Minor: it seems that all of these functions are the same. 

1. Calculate a summary only if required by the makeUrl command
2. Call makeUrl with either the values from the summary, or zeros and empty strings
3. Reassign the page URL.

If you were doing the entirely unimportant change mentioned above to pass in the handlers etc. you could pass in the same handler for all of them, and just pass that handler the arguments to the `makeUrl` function here. Or could pass a type to the handler, and it interpreted that type (retry test, skip test, restart tests etc.)

Just a thought.

modules/runner/src/main/ts/core/Callbacks.ts
```diff
@@ -0,0 +1,72 @@
+import { noop } from './Utils';
```
ephox-mogran, line 1: I think the separation of this file. Very clean.

modules/runner/src/main/ts/api/Main.ts
```diff
@@ -351,6 +127,13 @@ const loadtests = (data: HarnessResponse): void => {
 };
 
 const testrunner = (): void => {
+  const keepAliveTimer = setInterval(() => {
+    callbacks.sendKeepAlive(params.session, noop, () => {
+      // if the server shuts down stop trying to send messages
+      clearInterval(keepAliveTimer);
+    });
+  }, 5000);
```
ephox-mogran, line 135: The 5000 should probably be a constant or given some kind of description.

ephox-mogran: e.g. Use 5 seconds as a keep-alive heartbeat. Arbitrary chosen as 5. Or chosen to be 5 because some other value was x and y

modules/runner/src/main/ts/core/Reporter.ts
lnewson: Yeah, I really should have continued to split this out more. I got so far and was running out of time I allotted to work on this to be honest. I'll have another go and see how far I can get :+1:

modules/runner/src/main/ts/core/Callbacks.ts
lnewson: It'd be better if it didn't rely on a JQuery global (my one gripe) :stuck_out_tongue:

ephox-mogran: Yeah, requiring jquery is unfortunate. Is it mainly just for the ajax function now?

ephox-mogran: Are there any selectors beyond the capabilities of querySelector?

modules/runner/src/main/ts/core/Reporter.ts
ephox-mogran: I'll use `Really, really, really, really, really minor` next time so you feel less pressure to make changes to something that was already not in your sprint :)

modules/runner/src/main/ts/core/Utils.ts
lnewson: Yeah, runner needs to be able to work on IE 11 still.

modules/runner/src/main/ts/api/Main.ts
lnewson: Done

modules/runner/src/main/ts/core/Reporter.ts
lnewson: Okay, have done a fair bit more reworking. How does that look now?

modules/runner/src/main/ts/core/Utils.ts
lnewson: Done

modules/runner/src/main/ts/core/Reporter.ts
lnewson: Done, though in a bit of a different way...

modules/runner/src/main/ts/core/Callbacks.ts
lnewson: Doubtful, it likely could be rewritten entirely in native JS now.

lnewson:
I've reset the reviewers on this, as I made a lot more changes based on Morgan's feedback and also to try and break up the logic more.

modules/runner/src/main/ts/ui/Actions.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const Actions = (session: string): Actions => {
+  const reloadPage = (offset: number, failed: number, retry = 0) => {
+    const url = makeUrl(session, offset, failed, retry);
+    window.location.assign(url);
+  };
+
+  const updateHistory = (offset: number, failed: number, retry = 0) => {
+    window.history.pushState({}, '', makeUrl(session, offset, failed, retry));
```
ephox-mogran, line 18: Don't change it, but just curious if there is a reason you inlined this one but not the above one? Is there some particular heuristic you use?

modules/runner/src/test/ts/core/UtilsTest.ts
```diff
@@ -0,0 +1,45 @@
+import { assert } from 'chai';
```
ephox-mogran, line 1: Always good to see more tests :)

modules/runner/src/main/ts/ui/Actions.ts
lnewson: I just copied it across and that's how it was previously :shrug:

ephox-mogran: Fair enough.

lnewson: Fixed anyways :+1:

modules/runner/src/test/ts/reporter/ReporterTest.ts
```diff
@@ -0,0 +1,155 @@
...
+import { Reporter } from '../../../main/ts/reporter/Reporter';
+
+interface StartTestData {
+  session: string;
+  totalTests: number;
+  file: string;
+  name: string;
+}
+
+interface EndTestData {
```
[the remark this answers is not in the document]
lnewson: I keep forgetting to do that :disappointed: I went through and added it to all the types. Thanks for the reminder! :+1:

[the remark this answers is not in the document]
tinydylan: I wonder if there's an eslint rule for this.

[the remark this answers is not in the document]
lnewson: Only 2 rules I can see are the following and neither do what we want :disappointed: 
- https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-readonly-parameter-types.md
- https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-readonly.md

PR #25, opened 2020-06-30
Update bedrock dependencies

tinydylan:
It also needed the lockfile regenerated.


modules/server/package.json
```diff
@@ -18,33 +18,33 @@
...
     "recursive-readdir-sync": "^1.0.6",
-    "rollup": "^1.23.1",
-    "rollup-plugin-typescript2": "^0.24.0",
+    "rollup": "^2.18.1",
+    "rollup-plugin-typescript2": "^0.27.1",
     "serve-static": "^1.10.2",
-    "source-map-loader": "^0.2.3",
-    "ts-loader": "^5.3.3",
+    "source-map-loader": "^1.0.0",
+    "ts-loader": "^7.0.5",
```
lnewson, line 44: This can't be updated yet... last time I tried it broke a lot of things, so we need to be very careful about updating a number of these packages. This really needs an investigation/testing task assigned to it as well, given the potential it has for breaking things.

modules/server/package.json
```diff
@@ -18,33 +18,33 @@
...
     "tsconfig-paths-webpack-plugin": "^3.2.0",
     "typescript": "^3.3.0",
-    "webdriverio": "^5.14.5",
+    "webdriverio": "^6.1.22",
```
lnewson, line 47: This also worries me, as it may have breaking changes. Definitely needs more investigation, rather than just blindly upgrading everything.

ephox-mogran:
I know so little about our current tool chain and dependency list, that there isn't much point in reviewing. I'll remove myself.

tinydylan:
Will redo when time permits, with a less aggressive change and more testing.

PR #30, opened 2020-08-26
TINY-6218: Use glob to improve test directory scanning on windows

TheSpyder:
As a side effect of this, `--testdir` can now use globs to search multiple directories (on windows, *nix will still fail because the shell expands it). This happened because difference between `--testdir` and `--testdirs` is only in the command line parsing, they have identical implementations. I figured this was ok.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5113,11 +5113,6 @@ is-plain-object@^4.0.0:
   resolved "https://registry.yarnpkg.com/is-plain-object/-/is-plain-object-4.1.1.tgz#1a14d6452cbd50790edc7fdaa0aed5a40a35ebb5"
   integrity sha512-5Aw8LLVsDlZsETVMhoMXzqsXwQqr/0vlnBYzIXJbYo2F4yYlhLHs+Ez7Bod7IIQKWkJbJfxrWD7pA1Dw1TKrwA==
 
-is-promise@^2.1.0:
-  version "2.2.2"
-  resolved "https://registry.yarnpkg.com/is-promise/-/is-promise-2.2.2.tgz#39ab959ccbf9a774cf079f7b40c7a26f763135f1"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -5113,11 +5113,6 @@ is-plain-object@^4.0.0:
   resolved "https://registry.yarnpkg.com/is-plain-object/-/is-plain-object-4.1.1.tgz#1a14d6452cbd50790edc7fdaa0aed5a40a35ebb5"
   integrity sha512-5Aw8LLVsDlZsETVMhoMXzqsXwQqr/0vlnBYzIXJbYo2F4yYlhLHs+Ez7Bod7IIQKWkJbJfxrWD7pA1Dw1TKrwA==
 
-is-promise@^2.1.0:
-  version "2.2.2"
-  resolved "https://registry.yarnpkg.com/is-promise/-/is-promise-2.2.2.tgz#39ab959ccbf9a774cf079f7b40c7a26f763135f1"
-  integrity sha512-+lP4/6lKUBfQjZ2pdxThZvLUAafmZb8OAxFb8XXtiQmS35INgr85hdOGoEs124ez1FCnZJt6jau/T+alh58QFQ==
-
```
TheSpyder, line 5120: I'm not sure why this was removed, I just did `yarn workspace @ephox/bedrock-server add glob`. Maybe the lock file was out of date? I can revert this if we think it should remain.

lnewson: Nothing seems to have it listed as a dependency, so it was probably just yarn not cleaning up properly.

lnewson (approved):
Changelog entry please :smiley:

TheSpyder:
Oh yeah, changelogs are a thing :joy:

Do I need to bump the version as well? Or does CI do that?

ltrouton (approved):

PR #36, opened 2020-10-18
Added support for the new Chromium based Microsoft Edge browser

lnewson:
This adds support for the new Microsoft Edge browser, as it needs to load a different webdriver executable. I did wonder if we should add a new name for this, but it seemed to treat it more as an upgrade than a new browser.

### DIFF
--- a/modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
+++ b/modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
@@ -1,6 +1,8 @@
 import { ChildProcess } from 'child_process';
 import * as crossSpawn from 'cross-spawn';
 import * as http from 'http';
+import * as which from 'which';
+import * as Arr from '../util/Arr';
 
 export interface DriverAPI {
   start: (args?: string[]) => ChildProcess;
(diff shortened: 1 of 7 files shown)


modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
```diff
@@ -56,24 +70,29 @@ const loadPhantomJs = () => {
 };
 
 export const loadDriver = (browserName: string): DriverAPI => {
-  const driverDep = browserModules[browserName];
-  if (driverDep === undefined) {
+  const driverDeps = browserModules[browserName] || [];
```
ashfordneil, line 73: How much do we care about our regular tinymce coding style here in bedrock? I'd be inclined to suggest something like `Obj.getOpt(browserModules, browserName).getOr([])` but I don't know if that's appropriate here

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
```diff
@@ -56,24 +70,29 @@ const loadPhantomJs = () => {
...
   } else {
-    try {
-      return require(driverDep);
-    } catch (e) {
-      console.log(`No local ${driverDep} for ${browserName}. Searching system path...`);
+    const driver = findNpmPackage(driverDeps);
+    if (driver !== null) {
+      return driver;
+    } else {
+      console.log(`No local ${driverDeps[0]} for ${browserName}. Searching system path...`);
```
ashfordneil, line 81: [nitpick] should that be `${driverDeps.join(' / ')}` or something so that it shows we're definitely trying to locate the other drivers?

modules/server/src/main/ts/bedrock/util/Arr.ts
```diff
@@ -8,3 +8,14 @@ export const bind = <A, B> (xs: A[], f: (t: A) => B[]): B[] =>
 
 export const bind2 = <A, B, C> (xs: A[], f: (a: A) => B[], g: (b: B) => C[]): C[] =>
   bind(bind(xs, f), g);
+
+export const findMap = <A, B> (xs: ArrayLike<A>, f: (a: A, index: number) => B | null): B | null => {
+  for (let i = 0; i < xs.length; i++) {
+    const result = f(xs[i], i);
```
ashfordneil, line 14: Are we good to leave this untested? It's super basic, so it should be fine right?

ashfordneil (approved):

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
lnewson: We can't use any modules from the monorepo here, so we can't use `Optional`, `Obj`, etc... so we have to restrict the style a bit here.

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
lnewson: I wanted it to show the primary driver only here, as the others are fallbacks and shouldn't be expected to be used.

modules/server/src/main/ts/bedrock/util/Arr.ts
lnewson: It's basically a direct copy from `Arr` in the monorepo, just without `Optional` since we can't use that here. It might be worth adding a test though :shrug:

ltrouton (approved):

modules/server/src/main/ts/bedrock/util/Arr.ts
lnewson: As discussed we probably should, but it doesn't have tests atm so not worth the effort atm. Ideally, we should probably look at using katamari, since bedrock-server wouldn't have a circular dependency and therefore could actually use it.

metricjs (approved):

PR #37, opened 2020-10-29
Fixed IE 11 not shutting down with newer IEDriverServer versions

lnewson:
When using the latest version of 3.x for IEDriverServer, the `deleteSession()` call is being ignored when called too quickly, so we just need to introduce a small delay.

Have tested this branch using the driver from https://github.com/SeleniumHQ/selenium/tree/trunk/cpp/prebuilt/Win32/Release (3.105.1.1)

### DIFF
--- a/modules/server/src/main/ts/bedrock/auto/Driver.ts
+++ b/modules/server/src/main/ts/bedrock/auto/Driver.ts
@@ -141,7 +141,7 @@ const focusBrowser = (browserName: string, settings: DriverSettings) => {
   }
 };
 
-const setupShutdown = (driver: webdriverio.BrowserObject, driverApi: DriverLoader.DriverAPI): (immediate?: boolean) => Promise<void> => {
+const setupShutdown = (driver: webdriverio.BrowserObject, driverApi: DriverLoader.DriverAPI, shutdownDelay: number = 0): (immediate?: boolean) => Promise<void> => {
   const driverShutdown = (immediate?: boolean) => {
     try {
       if (immediate) {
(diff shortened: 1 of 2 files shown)


modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -141,15 +141,17 @@ const focusBrowser = (browserName: string, settings: DriverSettings) => {
   }
 };
 
-const setupShutdown = (driver: webdriverio.BrowserObject, driverApi: DriverLoader.DriverAPI): (immediate?: boolean) => Promise<void> => {
+const setupShutdown = (driver: webdriverio.BrowserObject, driverApi: DriverLoader.DriverAPI, shutdownDelay: number = 0): (immediate?: boolean) => Promise<void> => {
```
tinydylan, line 144: What's the interplay between `shutdownDelay` and `immediate`? Can we use the `immediate  behaviour when `shutdownDelay === 0` ?

tinydylan (approved):
Looks reasonable, just a question on the params.

modules/server/src/main/ts/bedrock/auto/Driver.ts
lnewson: It's more that this is different places calling it... as the delay is set up when creating the shutdown hook, whereas immediate is done via whatever tells the driver to shutdown.

As for `shutdownDelay === 0` we wouldn't want to do that, as `deleteSession` can take some time to run and we don't want to stop the driver until that has occurred by default.

lnewson: Looking more the immediate logic isn't used anymore, so it could actually be removed.

lnewson: For reference, the immediate logic is only used when the process is forcibly killed.

ltrouton (approved):

ashfordneil (approved):

PR #39, opened 2020-12-06
TINY-6732: Added BDD style tests using Mocha and changed bedrock to use Mocha predominately for it's runner

lnewson:
Description of changes:
- Adds support for mocha/BDD style tests (eg `it`, `describe`, etc...). This can either use the global or use an import from `bedrock-client` (we can tweak what we want to allow via linter rules).
- Adds support for skipping tests (to both the new and old-style tests).
- Changes the runner to primarily use `mocha` to run each test (keeps the current bedrock reporter/ui by using a custom mocha reporter). This opens up additional features from mocha, such as per test timeouts, retries, etc... and also allows us to run single tests by using `it.only`.
- Changes the runner to use Promises (via a polyfill) instead of callbacks.

One thing to note about this is that I've had to patch the mocha `it` functions to be able to:
- register the tests (so the server-generated imports can register the filenames)
- handle our custom test errors (since we allow strings, whereas Mocha only allows Errors).

Note: This still supports the existing bedrock chunking/hard reload logic. This is done by filtering out the tests to be omitted from mocha before starting the test run.


lnewson:
I should add I've exposed the `LoggedError` module so that we can still use additional log messages if needed (this is done by passing an `Error` and an array of log messages).

modules/client/src/main/ts/api/Bdd.ts
```diff
@@ -0,0 +1,36 @@
+import { Global } from '@ephox/bedrock-common';
+import { MochaGlobals } from 'mocha';
+
+const mochaGlobals: MochaGlobals = Global;
+
+const before = mochaGlobals.before;
```
techtangents, line 6: Would a destructuring assignment be nicer here?

modules/common/src/test/ts/api/ReporterTest.ts
```diff
@@ -1,9 +1,10 @@
 import { assert } from 'chai';
 import { describe, it } from 'mocha';
-import * as Reporter from '../../../main/ts/api/Reporter';
 import * as LoggedError from '../../../main/ts/api/LoggedError';
+import * as Reporter from '../../../main/ts/api/Reporter';
+import { HtmlDiffAssertionError } from '../../../main/ts/api/TestError';
 
-function htmlAssertion() {
+function htmlAssertion(): HtmlDiffAssertionError {
```
techtangents, line 7: arrow function?

modules/runner/src/main/ts/api/Main.ts
```diff
@@ -1,28 +1,36 @@
-import { Runner } from '../core/Runner';
-import { TestData } from '../core/TestTypes';
-import { Ui } from '../ui/Ui';
+import { Global } from '@ephox/bedrock-common';
+import * as Patches from '../core/Patches';
 import { UrlParams } from '../core/UrlParams';
 import { makeSessionId } from '../core/Utils';
 import { Callbacks } from '../reporter/Callbacks';
 import { Reporter } from '../reporter/Reporter';
+import { Runner } from '../runner/Runner';
```
techtangents, line 7: I was wondering if you were going to use something like this and replace the runner page: https://mochajs.org/#running-mocha-in-the-browser

Do you have some comments in the readme or somewhere on the approach to integrating mocha?

modules/runner/src/main/ts/reporter/MochaReporter.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+const defaultOptions: Partial<MochaReporterOptions> = {
+  onPass: noop,
+  onSkip: noop,
+  onFailure: noop
+};
+
+const isTestEqual = (test1?: Test, test2?: Test) => {
+  if (test1 === undefined || test2 === undefined) {
```
techtangents, line 25: boolean expression could be simplified. Not sure how it'd read, though.

techtangents: If both are undefined, should this return true?

techtangents (approved):
Nice.

modules/runner/src/main/ts/api/Main.ts
lnewson: That's effectively what's happening under the hood, we just don't use the mocha UI and keep our own. I wasn't keen to change this, as a lot of positioning tests might potentially fail since Mocha dumps a whole heap of content while the tests are running. Additionally, we would have lost the ability to have our "restart", "skip" or "retry" buttons.

modules/runner/src/main/ts/reporter/MochaReporter.ts
lnewson: In reality, neither should ever be `undefined` so I'll update the signature to account for that instead.

modules/client/src/main/ts/api/Bdd.ts
lnewson: It would be, however from a tree shaking perspective having them as separate variables would allow the unused variables to be dropped (have a tendency to do this because of that). Anyways, that doesn't matter here as we'd also need to mark it as side-effect free, so I'll go ahead change this to use object destructing instead.

modules/runner/src/main/ts/reporter/MochaReporter.ts
lnewson: Changed my mind... as it's potentially going to be more reusable/less error-prone if it accepts both. So have changed it to account for both being passed as undefined as well. I've also made the arguments non-optional, as that wasn't technically correct (we always want 2 arguments passed).

modules/common/src/test/ts/api/ReporterTest.ts
lnewson: Fixed

modules/client/src/main/ts/api/Bdd.ts
lnewson: Fixed

JamesToohey (approved):
Would this be a major version bump for Bedrock?

Can we add a test for .only (throw some errors in tests that should not run)?

Is it possible to have the amount of tests skipped reported in the command line (I had .only on this test):

Screen Shot 2020-12-07 at 2 42 30 pm

modules/common/src/main/ts/api/Failure.ts
```diff
@@ -0,0 +1,80 @@
...
+import * as TestLogs from './TestLogs';
+
+type TestLogs = TestLogs.TestLogs;
+type TestLogEntry = TestLogs.TestLogEntry;
+type TestError = TestError.TestError;
+type LoggedError = LoggedError.LoggedError;
+
+export type TestThrowable = TestLabel | TestError;
+
+const cleanStack = (error: Error, linesToRemove = 1) => {
```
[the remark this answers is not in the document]
lnewson: Untouched code, it was like this before and I'm not keen on changing it without having more details.

modules/runner/src/main/ts/reporter/MochaReporter.ts
```diff
@@ -0,0 +1,124 @@
...
+      afterEach(this: Context) {
+        const test = this.currentTest;
+        return lock.execute(() => processResult(test));
+      }
+    });
+
+    const onTestStart = (test: Test) => {
+      if (!isTestEqual(test, currentTest)) {
+        currentTest = test;
+        const testTitle = test.titlePath().slice(0, -1).join(' / ') + ' ' + test.title;
```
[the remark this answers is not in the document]
lnewson: It's not a string for starters, it's an array. So this removes the last entry from the array and then joins it back together so that we end up with something like this:
```
Parent / Current - Test
```

modules/sample/src/test/ts/client/SyncBddPassTest.ts
```diff
@@ -0,0 +1,49 @@
...
+      setTimeout(done, 100);
+    });
+
+    it('should work with retries', () => {
+      retryCount++;
+      assert.eq(3, retryCount);
+    }).retries(2);
+
+    it.skip('should be skipped with outer call', () => {
+      assert.eq(-1, count);
```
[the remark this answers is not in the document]
lnewson: I don't think it really matters? Though I guess that would be more obvious it should never hit that code :thinking:

modules/client/src/test/ts/core/TypeTest.ts
```diff
@@ -0,0 +1,30 @@
...
+    fc.assert(fc.property(fc.boolean(), (bool) => {
+      assert.equal(Type.typeOf(bool), 'boolean');
+    }));
+  });
+
+  it('is "object" for object', () => {
+    fc.assert(fc.property(fc.object(), (obj) => {
+      assert.equal(Type.typeOf(obj), 'object');
+    }));
+  });
```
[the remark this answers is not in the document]
lnewson: I only added a test for what the code covered directly tbh. I'll add one for a few others as well.

modules/runner/src/main/ts/reporter/MochaReporter.ts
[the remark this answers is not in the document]
JamesToohey: Oh yep, I misread this - all good

modules/sample/src/test/ts/client/SyncBddPassTest.ts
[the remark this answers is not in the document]
JamesToohey: Yeah that was why I suggested it. I read this and my reaction was "why are there assertions in here when this is never run?"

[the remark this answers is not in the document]
JamesToohey: Also since there's a `count--` there is a possible crazy chaos scenario where this test passes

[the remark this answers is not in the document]
lnewson: Fixed

modules/client/src/test/ts/core/TypeTest.ts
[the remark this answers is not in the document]
lnewson: Fixed

lnewson:
> Can we add a test for .only (throw some errors in tests that should not run)?

I've added something to test this as well. It needs a separate bedrock run though since only will kill all other tests.

lnewson:
> Is it possible to have the amount of tests skipped reported in the command line (I had .only on this test):

This does happen with these changes (you ran mocha, not bedrock in your example). Here's an example of what the output looks like:
image

lnewson:
> Would this be a major version bump for Bedrock?
Yeah, you'll see I've added the changelog entry for it. With bedrock we don't change the version number manually and change it using `lerna version` once changes have been merged.

ltrouton (approved):

PR #42, opened 2021-01-04
TINY-6732: Updated bedrock to support pulling in specified polyfills

lnewson:
This adds the ability to include additional polyfills as part of the test run by adding for example `--polyfills Promise Symbol`. The default is `Symbol` to keep backwards compatibility and I didn't add `Promise` as part of that as discussed with the Editor Team on slack.

This will allow us to write `async`/`await` tests in TinyMCE and rely on the Promise eslint rule to ensure we don't misuse a Promise for IE 11.

### DIFF
--- a/modules/server/src/main/ts/BedrockFramework.ts
+++ b/modules/server/src/main/ts/BedrockFramework.ts
@@ -23,17 +23,13 @@ export const go = (settings: BedrockFrameworkSettings): void => {
   }).then((driver) => {
     const webdriver = driver.webdriver;
     const serveSettings: Serve.ServeSettings = {
-      projectdir: settings.projectdir,
-      basedir: settings.basedir,
+      ...settings,
       driver: Attempt.passed(webdriver),
       testfiles: [],
       master,
       runner,
-      loglevel: settings.loglevel,
       customRoutes: undefined,
       stickyFirstSession: true,
-      overallTimeout: settings.overallTimeout,
-      singleTimeout: settings.singleTimeout,
       skipResetMousePosition: false
     };
 
--- a/modules/server/src/main/ts/BedrockFramework.ts
+++ b/modules/server/src/main/ts/BedrockFramework.ts
@@ -30,7 +30,7 @@ export const go = (settings: BedrockFrameworkSettings): void => {
       master,
       runner,
       loglevel: settings.loglevel,
-      customRoutes: settings.customRoutes,
+      customRoutes: undefined,
       stickyFirstSession: true,
       overallTimeout: settings.overallTimeout,
       singleTimeout: settings.singleTimeout,
--- a/modules/server/src/main/ts/BedrockFramework.ts
+++ b/modules/server/src/main/ts/BedrockFramework.ts
@@ -28,8 +28,10 @@ export const go = (settings: BedrockFrameworkSettings): void => {
       testfiles: [],
       master,
       runner,
+      // Framework mode doesn't support the --customRoutes argument, so just pass undefined
       customRoutes: undefined,
       stickyFirstSession: true,
+      // Framework mode doesn't support the --skipResetMousePosition argument, so never skip
       skipResetMousePosition: false
     };
 
--- a/modules/server/src/main/ts/BedrockManual.ts
+++ b/modules/server/src/main/ts/BedrockManual.ts
@@ -13,20 +13,11 @@ export const go = (settings: BedrockManualSettings): void => {
 
   routes.then((runner) => {
     const serveSettings: Webpack.WebpackServeSettings = {
-      projectdir: settings.projectdir,
-      basedir: settings.basedir,
-      testfiles: settings.testfiles,
+      ...settings,
       // There is no driver for manual mode.
       driver: Attempt.failed('There is no webdriver for manual mode'),
       master: null, // there is no need for master,
       runner,
-      loglevel: settings.loglevel,
-      customRoutes: settings.customRoutes,
-      config: settings.config,
-      coverage: settings.coverage,
-      overallTimeout: settings.overallTimeout,
-      singleTimeout: settings.singleTimeout,
-      polyfills: settings.polyfills,
       // sticky session is used by auto mode only
       stickyFirstSession: false,
       // reset mouse position will never work on manual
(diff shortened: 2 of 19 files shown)


modules/server/src/main/ts/BedrockManual.ts
```diff
@@ -26,6 +26,7 @@ export const go = (settings: BedrockSettings): void => {
       coverage: settings.coverage,
       overallTimeout: settings.overallTimeout,
       singleTimeout: settings.singleTimeout,
+      polyfills: settings.polyfills,
```
techtangents, line 29: Do we use all of `settings` here, or just a subset?

modules/server/src/main/ts/bedrock/compiler/Imports.ts
```diff
@@ -129,10 +142,15 @@ window.removeEventListener('error', importErrorHandler);
...
-export const generateImports = (useRequire: boolean, scratchFile: string, srcFiles: string[]): string => {
+export const generateImports = (useRequire: boolean, scratchFile: string, srcFiles: string[], polyfills: string[]): string => {
+  const polyfillPaths: Record<string, string> = {};
+  polyfills.forEach((name) => {
+    polyfillPaths[name] = convertPolyfillNameToPath(name);
+  });
+
   if (hasTs(srcFiles)) {
-    return generateImportsTs(useRequire, scratchFile, srcFiles);
+    return generateImportsTs(useRequire, scratchFile, srcFiles, polyfillPaths);
```
techtangents, line 152: Since both functions are called identically, you could refactor this. 

```
const f = hasTs(srcFiles) ? generateImportsTs : generateImportsJs;
f(useRequire, scratchFile, srcFiles, polyfillPaths);
```

modules/server/src/test/ts/compiler/ImportsTest.ts
```diff
@@ -0,0 +1,99 @@
+import { assert } from 'chai';
+import * as fc from 'fast-check';
+import { describe, it } from 'mocha';
+import { generateImports } from '../../../main/ts/bedrock/compiler/Imports';
+
+const validPolyfills = [ 'ArrayBuffer', 'Map', 'Object', 'Promise', 'Set', 'Symbol', 'TypedArray', 'WeakMap', 'WeakSet' ];
+
+const convertPolyfillName = (name: string) => {
```
techtangents, line 8: This seems to duplicate production code.

techtangents: I wonder if this whole test is a bit too much like the implementation. You may be a better off just using standard unit tests for this.

techtangents (approved):

techtangents:
Looks reasonable. I'm just querying a couple of very minor things on code style and tests - nothing blocking merge.

modules/server/src/main/ts/BedrockManual.ts
lnewson: All of it is used in various places of the code IIUC.

modules/server/src/main/ts/bedrock/compiler/Imports.ts
lnewson: Good point, I didn't even look at that since that part was done previously (I just was focused on passing the polyfills through).

modules/server/src/main/ts/BedrockManual.ts
techtangents: I was just wondering if we could use `...settings` in here.

modules/server/src/main/ts/bedrock/compiler/Imports.ts
techtangents: Yeah, it's an aside.

modules/server/src/test/ts/compiler/ImportsTest.ts
lnewson: Yeah, there's nothing testing that specific function atm, so should probably export that and unit test it. Also worth noting this isn't a duplicate, as the actual code determines the whole path while this only does the module name.

modules/server/src/main/ts/BedrockManual.ts
lnewson: Ahh I see, then probably... we'd still need the few extra things not in `settings`. Most of this is still hanging around from before we made the conversion to TypeScript, so it's probably just that we hadn't thought of switching to that.

I'll give it a shot and see how it goes anyways :shrug:

modules/server/src/test/ts/compiler/ImportsTest.ts
lnewson: Have added something to cover the polyfill name conversion as well. How does that look?

modules/server/src/main/ts/bedrock/compiler/Imports.ts
lnewson: Fixed

modules/server/src/main/ts/BedrockManual.ts
lnewson: Done. It does cause other settings to be pulled in, in some cases, but that should be an issue and makes it easier to add new settings.

modules/server/src/main/ts/BedrockFramework.ts
```diff
@@ -23,17 +23,13 @@ export const go = (settings: BedrockFrameworkSettings): void => {
...
-      projectdir: settings.projectdir,
-      basedir: settings.basedir,
+      ...settings,
       driver: Attempt.passed(webdriver),
       testfiles: [],
       master,
       runner,
-      loglevel: settings.loglevel,
-      customRoutes: settings.customRoutes,
+      customRoutes: undefined,
```
techtangents, line 31: This doesn't look right.

lnewson: This is correct, it's that the types were incorrect as `settings.customRoutes` was always `undefined` as it's not a valid setting for framework mode. See: https://github.com/tinymce/bedrock/blob/master/modules/server/src/main/ts/bedrock/cli/Clis.ts#L59

JamesToohey (approved):

modules/server/src/main/ts/BedrockFramework.ts
lnewson: As discussed on slack I've added a comment to state why we are passing undefined.

PR #43, opened 2021-01-06
Release Bedrock 10

lnewson:
Ther might be something else needed, but this should be everything required I believe. If something is missed there's really no reason we can't just release another version.


metricjs (approved):

JamesToohey (approved):

techtangents (approved):
This is just an aggregate of the last few PRs, right? It's a bit big to review as it is, so just rubberstamping. At this stage, any issues are going to be found using it in anger.

lnewson:
> This is just an aggregate of the last few PRs, right? It's a bit big to review as it is, so just rubberstamping. At this stage, any issues are going to be found using it in anger.

Yeah, that's correct.

lnewson:
Admin merging, as DeepScan seems to be having issues at the moment.

PR #46, opened 2021-01-12
Remove "framework" mode, as it's no longer in use.

techtangents:


modules/server/src/main/ts/bedrock/core/Settings.ts
```diff
@@ -8,38 +8,23 @@ export interface BedrockSettings {
...
-
-export interface BedrockManualSettings extends BedrockSettings {
   readonly config: string;
   readonly coverage: string[];
   readonly customRoutes: string;
   readonly polyfills: string[];
   readonly testfiles: string[];
 }
 
+export type BedrockManualSettings = BedrockSettings;
```
lnewson, line 18: I'd like for this to remain the same as `Clis`, so can we either revert this change or move all the options from `forManual` to `commonOptions`?

lnewson (approved):

modules/server/src/main/ts/bedrock/core/Settings.ts
techtangents: I've done the latter. I also removed the weird stopOnFailure__hidden setting. 

Can you please double-check the changes?

ashfordneil (approved):

spocke (approved):

PR #47, opened 2021-01-12
TINY-6882: Fixed beforeEach and afterEach hooks running in the wrong order

lnewson:
See the JIRA for more details and a replication case in Mocha to show this is how it should be working.

### DIFF
--- a/modules/runner/src/test/ts/runner/HooksTest.ts
+++ b/modules/runner/src/test/ts/runner/HooksTest.ts
@@ -55,15 +55,15 @@ describe('Hooks.runHooks', () => {
     RunnerTestUtils.populateHooks(nested, 2, (idx) => callStack.push('nested' + idx));
   });
 
-  it('runs beforeEach hooks from bottom up', () => {
+  it('runs beforeEach hooks from top down', () => {
     return Hooks.runHooks(nested, HookType.BeforeEach, true).then(() => {
-      assert.deepEqual(callStack, [ 'nested0', 'nested1', 'root0', 'root1' ]);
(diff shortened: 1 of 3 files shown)


modules/runner/src/test/ts/runner/HooksTest.ts
```diff
@@ -55,15 +55,15 @@ describe('Hooks.runHooks', () => {
...
-      assert.deepEqual(callStack, [ 'nested0', 'nested1', 'root0', 'root1' ]);
+      assert.deepEqual(callStack, [ 'root0', 'root1', 'nested0', 'nested1' ]);
     });
   });
 
-  it('runs afterEach hooks from top down', () => {
+  it('runs afterEach hooks from bottom up', () => {
     return Hooks.runHooks(nested, HookType.AfterEach, true).then(() => {
-      assert.deepEqual(callStack, [ 'root1', 'root0', 'nested1', 'nested0' ]);
+      assert.deepEqual(callStack, [ 'nested0', 'nested1', 'root0', 'root1' ]);
```
lnewson, line 66: I will say, this isn't the order I expected, as I would have thought it'd do `[ 'nested1', 'nested0', 'root1', 'root0' ]`, however the mocha test page confirmed this is how it works.

lnewson: Ahh, right it's because when writing these you'd naturally declare the `afterEach` hooks in reverse order within the same suite.

JamesToohey (approved):

metricjs (approved):

ltrouton (approved):

PR #48, opened 2021-01-18
TINY-6897: Fixed bedrock not returning a non-zero error code when tests unexpectedly stopped

lnewson:
Note: I haven't written any tests for the server part, as there are none. We're going to have to handle that tech debt one day though, so I'll log a JIRA so we can schedule it (see TINY-6898).


techtangents (approved):
As long as you cover this with a enough manual testing, that's fine.

ltrouton (approved):

ashfordneil (approved):

PR #50, opened 2021-02-23
TINY-6970: Add .mjs file support

TheSpyder:
We are in the process of switching to `.mjs` files in RTC in order to satisfy the NodeJS requirements for using a mixed ESM/CJS environment.

However Webpack applies a very strict `javascript/esm` type to `.mjs` files which make it much more difficult to import CommonJS modules. The error that it throws is `Can't import the named export <blah> from non EcmaScript module (only default export is available)`. That's not easy to google for, many have had this issue and it took a while to track down the best workaround. The solution is to redefine `.mjs` files as the more general `javascript/auto` type.

This PR makes a few changes:

* Adds `Test.mjs` to the CLI `testdir` extension search list
  * Side note: we can potentially remove `Test.bs.js` now but that extension might be useful for other ReScript projects
* Adjusts the generated import code to use the full file path (this won't break existing projects but is essential for `.mjs`)
* Adds `.mjs` as a supported Webpack file extension everywhere `.js` is available
* Refactors the webpack rules common to TS and JS projects
* Overrides the `.mjs` file type as `javascript/auto`

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,7 @@
+Version 11.1.0
+
+* Added support for `.mjs` test files, overridden to load with webpack's `javascript/auto` type instead of `javascript/esm`
+
 Version 11.0.2
 
 * Fixed a regression introduced in 11.0.0 whereby --customRoutes couldn't be used in manual mode.
--- a/modules/server/src/main/ts/bedrock/compiler/Webpack.ts
+++ b/modules/server/src/main/ts/bedrock/compiler/Webpack.ts
@@ -39,16 +39,15 @@ const webpackRemap: Array<Record<string, any>> = moduleAvailable('@ephox/swag') 
 
 const webpackSharedRules = webpackRemap.concat([
   {
-    test: /\.mjs$/,
+    test: /\.(mjs)$/,
     type: 'javascript/auto',
     use: []
   },
   {
-    test: /\.js|\.mjs$/,
+    test: /\.(js|mjs)$/,
     use: [ 'source-map-loader' ],
     enforce: 'pre'
   },
-
   {
     test: /\.(html|htm|css|bower|hex|rtf|xml|yml)$/,
     use: [ 'raw-loader' ]
@@ -92,7 +91,7 @@ const getWebPackConfigTs = (tsConfigFile: string, scratchFile: string, dest: str
     module: {
       rules: webpackSharedRules.concat([
         {
-          test: /\.tsx?$/,
+          test: /\.(tsx?)$/,
           use: [
             {
               loader: 'ts-loader',
@@ -114,7 +113,7 @@ const getWebPackConfigTs = (tsConfigFile: string, scratchFile: string, dest: str
       ]).concat(
         coverage ? [
           {
-            test: /\.tsx?$/,
+            test: /\.(tsx?)$/,
             enforce: 'post',
             loader: 'istanbul-instrumenter-loader',
             include: coverage.map((p) => path.resolve(p)),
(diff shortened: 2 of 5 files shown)


modules/server/src/test/ts/compiler/ImportsTest.ts
```diff
@@ -63,13 +63,38 @@ addTest("/${filename}.js");`
...
 __currentTestFile = "/${filename}.js";
-import "${filename}";
+import "${filename}.js";
 addTest("/${filename}.js");`
           );
         });
       });
     });
   });
+  context('ESM', () => {
```
lnewson, line 73:
```suggestion

  context('ESM', () => {
```
Newline for better readability

modules/server/src/main/ts/bedrock/compiler/Webpack.ts
```diff
@@ -32,11 +32,29 @@ const moduleAvailable = (name: string) => {
...
+    test: /\.mjs$/,
+    type: 'javascript/auto',
+    use: []
+  },
+  {
+    test: /\.js|\.mjs$/,
+    use: [ 'source-map-loader' ],
+    enforce: 'pre'
+  },
+
```
lnewson, line 51: Is there a reason we have a new line here, but not above?

lnewson (approved):
LGTM, though don't forget to add a changelog entry please.

- [ ] Add changelog entry

modules/server/src/main/ts/bedrock/compiler/Webpack.ts
```diff
@@ -32,11 +32,29 @@ const moduleAvailable = (name: string) => {
 
 const webpackRemap: Array<Record<string, any>> = moduleAvailable('@ephox/swag') ? [
   {
-    test: /\.js|\.tsx?$/,
+    test: /\.js|\.mjs|\.tsx?$/,
```
ashfordneil, line 35: I realise this isn't related to your changes, but should this be:
```suggestion
    test: /(\.js|\.mjs|\.tsx?)$/,
```
or even
```suggestion
    test: /\.(js|mjs|tsx?)$/,
```

So that things like `file.mjs-other-extension.html` don't match? It's out of scope, so feel free to ignore, but it looks like a few other regexes in this file have the same problem and it should be an easy fix if you want to do it.

ashfordneil (commented):
The code looks good, but I don't really understand enough about what's going on in the big picture to give it a tick confidently.

modules/server/src/main/ts/bedrock/compiler/Webpack.ts
TheSpyder: Either an oversight or a copy/paste error. I'll fix it.

modules/server/src/main/ts/bedrock/compiler/Webpack.ts
TheSpyder: ah, yeah that's probably what was intended.

modules/server/src/main/ts/bedrock/compiler/Webpack.ts
```diff
@@ -163,20 +170,10 @@ const getWebPackConfigJs = (scratchFile: string, dest: string, coverage: string[
...
-        {
-          test: /\.(html|htm|css|bower|hex|rtf|xml|yml)$/,
-          use: [ 'raw-loader' ]
-        }
-      ]).concat(
+      rules: webpackSharedRules.concat(
         coverage ? [
           {
-            test: /\.js?$/,
+            test: /\.js|\.mjs?$/,
```
TheSpyder, line 176: Oh this one is hilarious, it was previously accepting `.j` files 😂

JamesToohey (approved):

changelog.txt
```diff
@@ -1,3 +1,7 @@
+Version 11.0.3
```
lnewson, line 1: This is an "Added" so this should be a minor release (ie 11.1.0).

spocke (approved):

changelog.txt
TheSpyder: Fixed

PR #51, opened 2021-02-24
TINY-7069: Fixed setting timeouts at execution time not working

lnewson:
This adds a "private" onChange listener so that the runner can be notified when a runnables timeout is changed so that it can be reset (this is equivalent to what Mocha does when calling `timeout(ms)`).

Note: Have also converted the changelog format.


modules/runner/src/main/ts/runner/Run.ts
```diff
@@ -93,19 +94,25 @@ export const runWithTimeout = (runnable: Runnable, context: Context, defaultTime
     return run(runnable, context);
   } else {
     return new Promise((resolve, reject) => {
-      let timedOut = false;
-      const timer = setTimeout(() => {
-        timedOut = true;
-        reject(Failure.prepFailure(new Error(`Test ran too long - timeout of ${timeout}ms exceeded`)));
-      }, timeout);
+      const timer = Timer(() => {
+        cleanup();
```
spocke, line 98: There is a lot calling things before it's really defined here it works in TS land but I always find this hard to read. Also at this point the timer would already have finished so the timer.stop() call executed by the cleanup is really not needed. Maybe the timer could have a pass/fail kind of thing. So you hook it up with to callbacks instead one that is when the timer has finished and one if it finishes early by the call to stop. The unbind would still be an issue with timeouts but if it had listeners it could be avoided but then the timer would be a lot more complex.

Something like this:
const timer = Timer(() => reject(...), resolve, cleanup)

Then timer.stop() would call the resolve a timeout would call reject and cleanup would be called on both those actions.

spocke (approved):

modules/runner/src/main/ts/core/Runnable.ts
```diff
@@ -6,6 +6,15 @@ export const createRunnable = (title: string, fn: ExecuteFn | undefined): Runnab
   let timeout = -1;
   let slow = 75;
   let state = RunnableState.NotRun;
+  const listeners: Record<string, Set<(value: number) => void>> = {};
+
+  const fireChange = (type: string, value: number) => {
```
ashfordneil, line 11: Should type be `retries | slow | timeout` as per the `_onChange` method? And should that type be pulled out

ashfordneil (approved):

modules/runner/src/main/ts/runner/Run.ts
lnewson: Yeah, I admit I wasn't overly happy about this... however we can't do that because we have 3 different paths to a resolution here. We have:
1. The runnable function completes, in which case it should ignore the timer, `resolve` immediately and cancel the timer.
2. The runnable function throws an error, in which case it should ignore the timer, `reject` immediately and cancel the timer.
3. The timer finishes, in which case the runnable function result is ignored and `reject` immediately.

I'll have more of a think tomorrow and see if I can come up with something cleaner.

modules/runner/src/main/ts/core/Runnable.ts
lnewson: It should not, as I intended for this part to be generic/work for anything largely. We only need the types for the public API on what can be registered as listeners to make sure someone doesn't accidentally listen on the wrong thing.

modules/runner/src/main/ts/runner/Run.ts
lnewson: Okay, I think that should be a lot nicer now as we're no longer calling things before they are defined. How does that look @spocke?

CHANGELOG.md
```diff
@@ -0,0 +1,480 @@
...
+### Added
+- added IFrameTest.js to the samples
+- introduced '=>' symbol in selectors to represent inside an iframe's contents
+- added --delayExit command line option to keep browser open after finishing test run
+
+### Changed
+- updated key effects and mouse effects to handle dispatching within iframes
+- changed click from an ActionSequence to a WebElement action because Edge supports it
+
+
```
JamesToohey, line 463: Double newline here

CHANGELOG.md
```diff
@@ -0,0 +1,480 @@
...
+- generates reports on timeouts
+
+### Changed
+- uses getText instead of getInnerHtml to match new selenium webdriver APIs
+
+### Improved
+- improves test report format
+- improves experimental SauceLabs integration
+
+
```
JamesToohey, line 477: A whole bunch of newlines?

JamesToohey (approved):

CHANGELOG.md
lnewson: Also like this previously. I've fixed it.

ltrouton (approved):

PR #52, opened 2021-03-04
TINY-7039: Enable the mouse position reset logic on Microsoft Edge

lnewson:
This makes Chrome Edge match how Chrome works when running tests and also fixes some flakes we've been seeing lately.


spocke (approved):
Nice to there is a way to reset the mouse position with webdrivers.

JamesToohey (approved):

ltrouton (approved):

ashfordneil (approved):

PR #65, opened 2021-05-29
Added support to print coloured output in the terminal

lnewson:
This is something I've been wanting to do as a "pet" project for quite a long time, as the errors reported via the terminal can be very hard to read. More than happy to change the colours I've used if anyone has better suggestions.

Description of changes:
- Reworked the error printing logic to generate a normalized `ErrorData` object from the various `Error` types. From that the errors can be formatted in a consistent way.
- Updated the runner and server to pass through the new `ErrorData` object instead of just a plain text string when a test fails. This also could open up the ability to have different reporters in the future if we want/need it.
- Changed the legacy assertion module to throw an Assertion error. This is done so that we can highlight the diff, as previously it was all part of the single message.
- Formally deprecated the old legacy assertion module so we see not to use it in an IDE.

Screenshots:

Success:
image

Failure:
image

### DIFF
--- a/modules/common/src/main/ts/api/ErrorExtractor.ts
+++ b/modules/common/src/main/ts/api/ErrorExtractor.ts
@@ -1,0 +1,109 @@
+import * as Differ from './Differ';
+import * as TestError from './TestError';
+import * as LoggedError from './LoggedError';
+
+type LoggedError = LoggedError.LoggedError;
+
+type TestError = TestError.TestError;
+type PprintAssertionError = TestError.PprintAssertionError;
+type HtmlDiffAssertionError = TestError.HtmlDiffAssertionError;
+type AssertionError = TestError.AssertionError;
+
+export interface BasicErrorData {
+  readonly type: string;
+  readonly message: string;
+  readonly diff?: {
+    readonly expected: string;
+    readonly actual: string;
+    readonly comparison: string;
+  };
+}
+
+export interface ErrorData extends BasicErrorData {
+  readonly logs?: string;
+  readonly stack?: string;
+}
+
+const stringify = (e: any) => {
+  if (e === undefined) {
+    return 'undefined';
+  } else if (typeof e === 'string') {
+    return e;
+  } else {
+    return JSON.stringify(e);
+  }
+};
+
+const extractError = (err?: LoggedError): TestError =>
+  err === undefined ? new Error('no error given') : err;
+
+const extractStack = (e: TestError): string => {
+  if (e.stack) {
+    return e.stack.split('\n')
+      .filter((line) => line.indexOf('at') !== -1)
+      .join('\n');
+  } else {
+    return '';
+  }
+};
+
+const extractErrorData = (e: HtmlDiffAssertionError | PprintAssertionError): BasicErrorData => {
+  const actual = e.diff.actual;
+  const expected = e.diff.expected;
+  const comparison = TestError.isHTMLDiffError(e) ? e.diff.comparison : Differ.diffPrettyText(actual, expected);
+  return ({
+    type: e.name,
+    message: `Test failure: ${e.message}`,
+    diff: {
+      actual,
+      expected,
+      comparison
+    }
+  });
+};
+
+const extractAssertionErrorData = (e: AssertionError): BasicErrorData => {
+  const actual = stringify(e.actual);
+  const expected = stringify(e.expected);
+  const message = `Assertion error: ${e.message ?? ''}`;
+  if (e.showDiff !== false) {
+    return {
+      type: e.name,
+      message,
+      diff: {
+        expected,
+        actual,
+        comparison: Differ.diffPrettyText(actual, expected)
+      }
+    };
+  } else {
+    return { type: e.name, message };
+  }
+};
+
+export const getBasicErrorData = (e: TestError): BasicErrorData => {
+  if (TestError.isHTMLDiffError(e) || TestError.isPprintAssertionError(e)) {
+    return extractErrorData(e);
+  } else if (TestError.isAssertionError(e)) {
+    return extractAssertionErrorData(e);
+  } else if (e.name && e.message) {
+    return { type: 'Error', message: e.name + ': ' + e.message };
+  } else if (e.toString !== undefined) {
+    return { type: 'Error', message: String(e) };
+  } else {
+    return { type: 'Error', message: JSON.stringify(e) };
+  }
+};
+
+export const getErrorData = (err: LoggedError): ErrorData => {
+  const e = extractError(err);
+  const formattedLogs = err.logs && err.logs.length > 0 ? err.logs.map((log) => log
+    .replace(/\n/g, '\\n')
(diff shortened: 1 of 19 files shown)


JamesToohey (approved):

modules/common/src/main/ts/api/ErrorExtractor.ts
```diff
@@ -0,0 +1,109 @@
...
+      expected,
+      comparison
+    }
+  });
+};
+
+const extractAssertionErrorData = (e: AssertionError): BasicErrorData => {
+  const actual = stringify(e.actual);
+  const expected = stringify(e.expected);
+  const message = `Assertion error: ${e.message ? e.message : ''}`;
```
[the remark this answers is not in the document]
lnewson: We're not guarnteed it'll be a string here and it could be `undefined` IIRC (this was like this previously and I seem to recall running into it at some stage).

modules/common/src/main/ts/api/ErrorExtractor.ts
```diff
@@ -0,0 +1,109 @@
...
+    readonly comparison: string;
+  };
+}
+
+export interface ErrorData extends BasicErrorData {
+  readonly logs?: string;
+  readonly stack?: string;
+}
+
+const stringify = (e: any) => {
```
ashfordneil, line 27:
```suggestion
const stringify = (e: unknown) => {
```

modules/common/src/main/ts/api/ErrorExtractor.ts
```diff
@@ -0,0 +1,109 @@
...
+  } else {
+    return JSON.stringify(e);
+  }
+};
+
+const extractError = (err?: LoggedError): TestError =>
+  err === undefined ? new Error('no error given') : err;
+
+const extractStack = (e: TestError): string => {
+  if (e.stack) {
```
ashfordneil, line 41:
```suggestion
  if (e.stack !== undefined) {
```

modules/common/src/main/ts/api/ErrorExtractor.ts
```diff
@@ -0,0 +1,109 @@
...
+    return { type: 'Error', message: String(e) };
+  } else {
+    return { type: 'Error', message: JSON.stringify(e) };
+  }
+};
+
+export const getErrorData = (err: LoggedError): ErrorData => {
+  const e = extractError(err);
+  const formattedLogs = err.logs && err.logs.length > 0 ? err.logs.map((log) => log
+    .replace(/\n/g, '\\n')
```
ashfordneil, line 101: Should this be a `JSON.stringify(log)` maybe? There might be some benefit to either always printing the string as itself (with actual newline characters in it) or always printing it as a valid string literal (with everything escaped in the same ways) instead of escaping only some characters.

Or from another perspective, if we're going to say `\n` => newline, shouldn't we be making sure that any existing `\` in the string becomes `\\` (to prevent an accidental `\n`)?

modules/common/src/main/ts/api/Reporter.ts
```diff
@@ -30,111 +22,72 @@ const processQUnit = (html: string): string =>
...
+  ErrorExtractor.getErrorData(err);
+
+export const dataHtml = (err: ErrorData): string =>
+  pprintError(err, htmlentities, (actual, expected, comparison) => {
+    return comparison !== undefined ? processQUnit(comparison) : Differ.diffPrettyHtml(actual, expected);
+  });
+
+export const dataText = (err: ErrorData): string =>
+  pprintError(err, identity, (actual, expected, comparison) => {
+    // TODO: get rid of the <ins> and <del> in the comparison. Probably need to change the code that throws HtmlAssertionError.
```
ashfordneil, line 81: Does this need a Jira?

modules/server/src/main/ts/bedrock/core/ConsoleReporter.ts
```diff
@@ -1,9 +1,45 @@
-import { TestResults } from '../server/Controller';
+import * as chalk from 'chalk';
+import { highlight } from 'cli-highlight';
+import { TestErrorData, TestResults } from '../server/Controller';
 
 const numberOfErrorsToPrint = 5;
 
-export const generateReport = (data: TestResults): string => {
+const highlightDiff = (diff: string) =>
+  highlight(diff, { language: 'diff' });
```
ashfordneil, line 8: Is this `cli-highlight` library an arbitrary syntax highlighter for all languages? Are we okay with pulling that in

ashfordneil (approved):

modules/common/src/main/ts/api/ErrorExtractor.ts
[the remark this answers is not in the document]
ashfordneil: I'm not good with the syntax, but is this a case for:
```suggestion
  const message = `Assertion error: ${e.message ?? ''}`;
```

modules/common/src/main/ts/api/ErrorExtractor.ts
lnewson: No, it's meant to be printed as plain text not an array. Either way, this was just moved code and it did the same previously.

modules/common/src/main/ts/api/Reporter.ts
lnewson: Moved code and was like this previously. So out of scope, however in saying that IMO it's not important as we are moving away from all the old assertion libraries and this will eventually become irrelvant.

modules/common/src/main/ts/api/ErrorExtractor.ts
lnewson: That would be a breaking change, so we can't do that. It's possible the stack could be any falsy variant (it could be undefined, could be an empty string, etc...). We don't have our nice Katamari APIs here so we still have to do this occasionally (it was also like this in the previous code before I moved it).

modules/common/src/main/ts/api/ErrorExtractor.ts
lnewson: `any` is the correct type here as per `JSON.stringify` types.

modules/server/src/main/ts/bedrock/core/ConsoleReporter.ts
lnewson: No idea what you're asking here sorry?

modules/common/src/main/ts/api/ErrorExtractor.ts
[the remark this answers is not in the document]
lnewson: Yeah, that seems like a valid use case so we probably should use it.

modules/server/src/main/ts/bedrock/core/ConsoleReporter.ts
lnewson: Discussed on slack. There are not many viable alternatives for this, so we're okay with this. If anyone finds anything, I'm more than happy to use something else.

ltrouton (approved):

PR #67, opened 2021-06-09
TINY-7578: Fixed hooks failing to register when passing undefined as the title

lnewson:


modules/runner/src/main/ts/core/Globals.ts
```diff
@@ -19,16 +19,11 @@ const getCurrentSuiteOrDie = (): Suite => {
...
-  if (typeof title === 'string' && fn !== undefined) {
-    const hook = Hook.createHook(`${type}: ${title}`, fn);
-    suite.hooks[type].push(hook);
-  } else if (typeof title === 'function') {
-    const hook = Hook.createHook(type, title);
-    suite.hooks[type].push(hook);
-  } else {
-    console.error('Failed to add hook', title, fn);
-  }
+const addHook = (suite: Suite, type: HookType, title: string | ExecuteFn | undefined, fn?: ExecuteFn) => {
```
ashfordneil, line 22: I know it doesn't type-check as well, but would this be clearer if we wrote it out with overloads?

```suggestion
const addHook: {
  (suite: Suite, type: HookType, title: string, fn: ExecuteFn) => void;
  (suite: Suite, type: HookType, title: undefined, fn: ExecuteFn) => void;
  (suite: Suite, type: HookType, fn: ExecuteFn) => void;
} = (suite, type, title, fn) => {
```

modules/runner/src/main/ts/core/Hook.ts
```diff
@@ -1,6 +1,6 @@
 import { ExecuteFn, Hook } from '@ephox/bedrock-common';
 import * as Runnable from './Runnable';
 
-export const createHook = (title: string, fn: ExecuteFn): Hook => {
+export const createHook = (title: string, fn: ExecuteFn | undefined): Hook => {
```
ashfordneil, line 4: Shouldn't this be the other way around?
```suggestion
export const createHook = (title: string | undefined, fn: ExecuteFn): Hook => {
```

ashfordneil (approved):

modules/runner/src/main/ts/core/Hook.ts
lnewson: No, this is the internal implementation not the public one. A title is mandatory for running and you'll see it's normalized to this in `Globals`.

modules/runner/src/main/ts/core/Globals.ts
lnewson: I had considered that, but it's pretty messy. Also worth noting the above isn't quite correct either. If we did the overload it'd be this:
```
const addHook: {
  (suite: Suite, type: HookType, title: string | undefined, fn: ExecuteFn) => void;
  (suite: Suite, type: HookType, fn: ExecuteFn) => void;
   (suite: Suite, type: HookType, title: string) => void;
} = (suite, type, title, fn) => {
```
because Mocha is weird and allows registering a hook without an actual function :shrug:

lnewson: Actually, correction, as both of the above is invalid and won't compile. It'd need to be this:
```
const addHook: {
  (suite: Suite, type: HookType, title: string, fn: ExecuteFn): void;
  (suite: Suite, type: HookType, title: undefined, fn: ExecuteFn): void;
  (suite: Suite, type: HookType, fn: ExecuteFn): void;
} = (suite: Suite, type: HookType, title: string | ExecuteFn | undefined, fn?: ExecuteFn) => {
  const hookTitle = typeof title === 'string' ? `${type}: ${title}` : type;
  const hookFn = typeof title === 'function' ? title : fn;
  const hook = Hook.createHook(hookTitle, hookFn);
  suite.hooks[type].push(hook);
};
```
I guess that does help to show what it's trying to handle.

lnewson: We'd then also have to add overloads in all the `before`, `after`, etc... functions. It's just really not a clean approach IMO.

TheSpyder: The type can be made an interface, and then used everywhere - that's how `@types/mocha` does it
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/618f10bd9a7440385a52003209dadcd1bcfe3e9b/types/mocha/index.d.ts#L385-L415

lnewson: We already do that here: https://github.com/tinymce/bedrock/blob/master/modules/common/src/main/ts/api/TestTypes.ts#L39 However, that is a "sane" set of types and doesn't allow for the hacks that `bs-mocha` is doing. I definitely don't want us to have types that allow `(title: undefined, fn: Function)`.

lnewson: Also worth noting even if you do the overload, you still need types on the actual implementation arguments because otherwise they are implicitly any. As an example, if we do this:
```
export const afterEach: HookFn = (title, fn?) => {
```
then we get the following compiler errors:
```
src/main/ts/core/Globals.ts(70,35): error TS7006: Parameter 'title' implicitly has an 'any' type.
src/main/ts/core/Globals.ts(70,42): error TS7006: Parameter 'fn' implicitly has an 'any' type.
```

JamesToohey (approved):

modules/runner/src/main/ts/core/Globals.ts
lnewson: For now, I've extracted the title types out into a separate type as it makes it a little more readable I think.

lnewson: Discussed on slack and we're just going to go with a comment to help clarify what's going on.

PR #74, opened 2021-08-28
TINY-7651: Upgrade to latest ts-loader, webpack and webdriverio

lnewson:
This is something I've been wanting to do for quite a while now, and can finally be done since we've removed the circular project reference in the TinyMCE repo. This PR makes the following changes:
- Updated to latest `ts-loader` (9.x) to get proper project reference support.
- Upgrade to latest `webpack` (5.x). This was needed to upgrade to latest ts-loader.
- Upgrade to latest `fork-ts-checker-webpack-plugin`, again to get proper project reference support.
- Upgrade to latest `webdriverio` version and set minimum Node.js version to match (>= 12.x)
- Swap to using async/await syntax as it has better webdriverio support (the types didn't play well when using .then).
- Upgrade to latest typescript version.
- Also includes some very minor linter fixes in the sample project.

Note: This includes breaking changes, so will require releasing Bedrock v12.


JamesToohey (approved):

modules/server/src/main/ts/bedrock/server/EffectUtils.ts
```diff
@@ -1,71 +1,64 @@
...
-      return true;
-    }).catch((e) => {
-      if (!(e.name && e.name === 'no such frame')) {
-        throw e;
-      }
-      return false;
-    });
-  };
+// eslint-disable-next-line @typescript-eslint/ban-types
+const frameSelected = (driver: Browser<'async'>, frame: object) => async (): Promise<boolean> => {
```
ashfordneil, line 9: Is it possible to be more specific with the type here, and indicate specifically that it needs to be the web driver element type (based on my quick scan of the docs)?

ashfordneil (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Improved
+- Upgrade to `ts-loader` 9.x and `fork-ts-checker-webpack-plugin` 6.x to properly support TypeScript project references.
+
+### Changed
+- Upgrade to `webdriverio` 7.x.
+- Upgrade to `webpack` 5.x.
+- Convert to using `async`/`await` syntax.
```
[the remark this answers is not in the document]
lnewson: No, it's just it's internals.

modules/server/src/main/ts/bedrock/server/EffectUtils.ts
lnewson: I have no idea what you mean here sorry...

lnewson: Oh, are you talking above `object` type for `frame`? That's unchanged here so that'd really be out of scope.

ashfordneil: Yeah the `object` type. I think it was `string` previously?

lnewson: Fixed... I remember why I did this now,  it's because I was spiking and the type isn't exposed, as it's an `ElementReference`, not an `Element` as you mentioned.  As such, we need to get the type using `ReturnType`.

CHANGELOG.md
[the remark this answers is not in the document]
lnewson:
```suggestion
- Convert to using `async`/`await` syntax on the server.
```

PR #75, opened 2021-08-28
TINY-7918: Improved the test execution speed on Chrome and Edge

lnewson:
This swaps the action used to reset the mouse position between tests on Chrome and Edge. For those that don't know the history, chromium browsers like to fire mouseover, etc... events for where the actual mouse, even when it's not moved, which can cause test flakes.

This version is ~3x faster than the old version (~10ms vs ~30ms), so while that's not much for a single test, when you have 5000+ tests as we do in TinyMCE, then it saves ~1&half; minutes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 11.5.0 - 2021-08-30
+
+### Improved
+- Improved test execution time on `Chrome` and `Edge` by using a faster mouse reset action.
+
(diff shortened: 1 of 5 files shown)


lnewson:
I should add the primary reason this is ~3x faster is because the old logic required 4 webdriver calls: lookup element, find position, find page scroll and move to element. So we're reducing the calls from being made from 4 -> 1.

lnewson:
Also just to show some performance stats, using the current version locally my tests took 371.71s to run. Using this newer version, it took 264.63s to run.

TheSpyder (approved):

JamesToohey (approved):

PR #78, opened 2021-09-09
TINY-7971: Added node environment definitions to webpack in JS mode

TheSpyder:
RTC runs a dual NodeJS + Bedrock test setup and some of the NodeJS-isms are too tightly integrated to remove easily.

This wasn't a problem with Webpack 4, it auto-bundles a lot of polyfills, but Webpack 5 stopped doing that.

I can manage `assert` and `util` by adding the webpack 4 versions as RTC dependencies but these polyfills still expect a `process` global. The only option I've found to mange code that uses `process.env` is a `DefinePlugin` (which does string replacement on bundled files).

I only did this to JS mode; I figure TypeScript code is less likely to need them.

### DIFF
--- a/modules/server/src/main/ts/bedrock/compiler/Webpack.ts
+++ b/modules/server/src/main/ts/bedrock/compiler/Webpack.ts
@@ -204,6 +204,12 @@ const getWebPackConfigJs = (scratchFile: string, dest: string, coverage: string[
         ] : []
       )
     },
+    plugins: [
+      new webpack.DefinePlugin({
+        'process.env.NODE_ENV': JSON.stringify('development'),
+        'process.env.NODE_DEBUG': JSON.stringify(false),
+      })
+    ],
     output: {
       filename: path.basename(dest),
       path: path.resolve(path.dirname(dest))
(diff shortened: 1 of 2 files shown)


modules/server/src/main/ts/bedrock/compiler/Webpack.ts
```diff
@@ -204,6 +204,12 @@ const getWebPackConfigJs = (scratchFile: string, dest: string, coverage: string[
         ] : []
       )
     },
+    plugins: [
+      new webpack.DefinePlugin({
```
lnewson, line 208: Should we be applying this in TS mode as well? I realise you only need it for JS mode, but I'm not too keen on diverging too much between the 2 modes.

lnewson (approved):

modules/server/src/main/ts/bedrock/compiler/Webpack.ts
TheSpyder: JS mode is really only used by RTC, so I was aiming to make it as small a change as possible, but I can add it to both.

lnewson: I just figured it shouldn't hurt anything to have it enabled by default anyways :shrug:

JamesToohey (approved):

TheSpyder:
After discussing the bedrock/mocha situation with Spocke last week I realised there was in fact a way to avoid depending on `assert` and `util` in RTC's bedrock tests 🎉

We still need the `NODE_ENV` replacement, because `immer` uses it, but we don't need `NODE_DEBUG`.

PR #80, opened 2021-11-17
TINY-7654: Use inline source maps for manual mode and ignore more changes

lnewson:
This just makes some changes to improve inconveniences I found when getting this working in TinyMCE:
- Bedrock manual mode will now use inline source maps. They take about the same time to generate but ensures the source maps don't run into caching issues or similar when running in watch mode.
- Adds a few more paths to ignore to prevent webpack doing multiple compilations when not necessary.

### DIFF
--- a/modules/server/src/main/ts/bedrock/compiler/Webpack.ts
+++ b/modules/server/src/main/ts/bedrock/compiler/Webpack.ts
@@ -157,7 +157,7 @@ const getWebPackConfigTs = (tsConfigFile: string, scratchFile: string, dest: str
           /\.(js|ts)\.map$/,
           // Something seems to trigger that node module package.json files change when they
           // haven't, so lets just ignore them entirely
-          /node_modules\/.*package\.json$/
+          /node_modules\/.*\/package\.json$/
         ]
       }),
       new webpack.DefinePlugin({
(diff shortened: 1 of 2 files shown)


modules/server/src/main/ts/bedrock/compiler/Webpack.ts
```diff
@@ -150,9 +150,15 @@ const getWebPackConfigTs = (tsConfigFile: string, scratchFile: string, dest: str
...
-      // See https://github.com/TypeStrong/ts-loader#usage-with-webpack-watch
       new webpack.WatchIgnorePlugin({
-        paths: [ /\.d\.ts$/ ]
+        paths: [
+          // Ignore generated files. See https://github.com/TypeStrong/ts-loader#usage-with-webpack-watch
+          /\.d\.ts$/,
+          /\.(js|ts)\.map$/,
+          // Something seems to trigger that node module package.json files change when they
+          // haven't, so lets just ignore them entirely
+          /node_modules\/.*package\.json$/
```
spocke, line 160: Super picky but should this be `/node_modules\/.*\/package\.json$/` so that it only matches `package.json` not `foo-package.json` files.

spocke (approved):

modules/server/src/main/ts/bedrock/compiler/Webpack.ts
lnewson: Oops... I had fixed that in one place (TinyMCE) but not the other it seems :cry: Thanks for the catch Spocke!

lnewson:
```suggestion
          /node_modules\/.*\/package\.json$/
```

ashfordneil (approved):

ltrouton (approved):

PR #81, opened 2021-12-05
TINY-8293: Swap from using the `@ephox/wrap-promise-polyfill` wrapper to using `promise-polyfill` directly

lnewson:


ltrouton (approved):

TheSpyder (approved):
Would be nice if we didn't need this at all, but I guess it doesn't hurt and means we can continue to upgrade bedrock in TinyMCE 5.

lnewson:
> Would be nice if we didn't need this at all, but I guess it doesn't hurt and means we can continue to upgrade bedrock in TinyMCE 5.

Yeah being able to continue to use it for older versions was the main reason I didn't just rip it out, as that could be useful. I suspect we'll get to a point where we may want to use async/await or similar but for now it's almost no extra cost for us to keep the runner working on older browsers.

ashfordneil (approved):

PR #83, opened 2021-12-24
TINY-8403: Fixed the server not always getting shutdown correctly

lnewson:
This was something I accidentally broke in 12.0.0 via https://github.com/tinymce/bedrock/pull/74. The problem is if a successful shutdown happens the `driverApi.stop()` function wasn't being called. That in turn meant the webdriver process would keep running forever depending on the shell/terminal being used. Additionally, if an unexpected runner error occurred it wasn't doing a graceful server shutdown and instead did a forced shutdown (since the `Reporter.writePollExit` call returns a rejected promise).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


ltrouton (approved):

lnewson:

PR #84, opened 2021-12-25
TINY-8189: Fixed the wrong timeout value being reported

lnewson:
If a bedrock configured/default timeout occurred, then the wrong timeout value was reported in the error messages. This means you would have ended up with something like `Error: Test ran too long - timeout of -1ms exceeded`, so this ensures the correct timeout value is set for the runnable as that'll make sure if it's looked up in the test it has the right value as well.

Note: Also fixed a minor issue with the URL being shown incorrectly based on the state when a test is retried and then passes.

### DIFF
--- a/modules/runner/src/main/ts/runner/Runner.ts
+++ b/modules/runner/src/main/ts/runner/Runner.ts
@@ -42,7 +42,7 @@ export const Runner = (rootSuite: Suite, params: UrlParams, callbacks: Callbacks
     const sum = reporter.summary();
     if (params.retry > 0) {
       params.retry = 0;
-      actions.updateHistory(params.offset, params.retry, sum.skipped);
+      actions.updateHistory(params.offset, sum.failed, sum.skipped, params.retry);
     }
   };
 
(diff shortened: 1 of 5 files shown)


modules/runner/src/main/ts/runner/Runner.ts
```diff
@@ -42,7 +42,7 @@ export const Runner = (rootSuite: Suite, params: UrlParams, callbacks: Callbacks
     const sum = reporter.summary();
     if (params.retry > 0) {
       params.retry = 0;
-      actions.updateHistory(params.offset, params.retry, sum.skipped);
+      actions.updateHistory(params.offset, sum.failed, params.retry, sum.skipped);
```
ltrouton, line 45: It looks like the interface for the `updateHistory` fn is the following. I am wondering if there is an existing mistake somewhere since the `skipped` and `retry` arguments don't look to be in the right order? I could be missing something obvious though.
```
readonly updateHistory: (offset: number, failed: number, skipped: number, retry?: number) => void;
```

ltrouton (approved):

modules/runner/src/main/ts/runner/Runner.ts
lnewson: Good catch thanks! I only noticed the failure argument was missing so I just added it, but you're right this had a few things wrong.

lnewson:

PR #85, opened 2021-12-28
TINY-8338: Upgrade to latest webpack-dev-server

lnewson:
Description of changes:
- Upgrade to webpack-dev-server 4.x
- Upgrade other dev dependencies to the latest version (including TypeScript 4.5.x)
- Removed the legacy/deprecated assert API from the client


modules/server/src/main/ts/bedrock/compiler/Webpack.ts
```diff
@@ -296,52 +296,70 @@ export const devserver = async (settings: WebpackServeSettings): Promise<Serve.S
   const tsConfigFile = settings.config;
 
   const compileInfo = await getCompileInfo(tsConfigFile, scratchDir, settings.basedir, true, settings.testfiles, settings.coverage);
-  return Serve.startCustom(settings, (handler) => {
+  return Serve.startCustom(settings, (port, handler) => {
     const scratchFile = compileInfo.scratchFile;
-    console.log(`Loading ${settings.testfiles.length} tests...`);
+    console.log(`Loading ${settings.testfiles.length} test files...`);
```
lnewson, line 301: Updated this console message as there's normally a few tests per file so this was misleading.

ltrouton (approved):

modules/runner/src/main/ts/runner/Run.ts
```diff
@@ -94,9 +94,13 @@ const runWithCleanup = (runnable: Runnable, context: Context, cleanup: () => voi
   });
 
 export const runWithTimeout = (runnable: Runnable, context: Context, defaultTimeout: number): Promise<void> => {
+  // Update the runnable timeout to use the default timeout if required
```
spocke, line 97: We finally got around fixing this one. :+1:

modules/server/src/main/ts/bedrock/server/Serve.ts
```diff
@@ -86,31 +85,37 @@ export const startCustom = async (settings: ServeSettings, createServer: (listen
...
 
 export const start = (settings: ServeSettings): Promise<ServeService> => {
-  return startCustom(settings, http.createServer);
+  return startCustom(settings, (port, listener) => {
+    const server = http.createServer(listener);
+    return {
+      start: () => {
+        server.listen(port);
+        return Promise.resolve();
+      },
```
spocke, line 113: Not sure how important this is since setting up a listening socket is probably very fast. But it feels more correct to use the async callback for this one that just resolving it.

```suggestion
      start: () => {
        return new Promise((resolve) => {
          server.listen(port, resolve);
        });
      },
```

spocke (approved):

modules/server/src/main/ts/bedrock/server/Serve.ts
lnewson: Good catch, thanks Spocke!

lnewson: Worth noting we weren't waiting for this previously either 😢

PR #86, opened 2021-12-28
TINY-8165: Fixed cross frame uncaught errors not handled correctly

lnewson:
This fixes uncaught errors thrown within frames not getting their stack trace used due to the `instanceof Error` check failing. I've moved the `Type` module to bedrock common since it's now used by the client and runner (remember we can't use katamari).


ltrouton (approved):

lnewson:

PR #98, opened 2022-05-10
TINY-8705: Adding extraBrowserCapabilities flag to bedrock-auto

ephox-mogran:


modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -67,6 +68,14 @@ const addArguments = (capabilities: Record<string, any>, name: string, args: str
   capabilities[name].args = currentArgs.concat(args);
 };
 
+const getExtraBrowserCapabilities = (settings: DriverSettings): string[] => {
+  if (settings.extraBrowserCapabilities && settings.extraBrowserCapabilities.length > 0) {
```
spocke, line 72: Will this ever be undefined? Doesn't the default value of the ClOptions set this as an empty string? If it can be undefined then it should be marked with a ? in the types.

modules/server/src/test/ts/ClisTest.ts
```diff
@@ -94,11 +95,68 @@ describe('Clis.forAuto', () => {
...
+      bucket: 1,
+      buckets: 1,
+      useSandboxForHeadless: false,
+      extraBrowserCapabilities: ' --some-browser-flag',
+      skipResetMousePosition: false,
+      wipeBrowserCache: false
+    }, cleanResult(actual));
+  });
+
+  // This isn't so much of desired test outcome, just a reality. If we don't have
```
spocke, line 141: Will it work with quoted strings like `--foo "--bar"` or does it need to be prefixed by a space. That will be very hard to spot.

spocke (approved):

modules/server/src/main/ts/bedrock/auto/Driver.ts
ephox-mogran: I'll test if it is defaulting it properly. I tried a few things to get around the -- issue, and this was probably an artifact of that.

- [ ] see if I can remove the falsy check.

modules/server/src/test/ts/ClisTest.ts
ephox-mogran: Yeah, it's a compromise. The `command-line-arguments` that bedrock is currently using just uses regex to identify what everything is. Anything that starts with a --, even if it has quotes surrounding it, is considered an option.  The quotes just make it a single value to `argv`. It doesn't need to be a space ... just anything, because the code it is trying to bypass is this: 

```
const re = {
  short: /^-([^\d-])$/,
  long: /^--(\S+)/,
  combinedShort: /^-[^\d-]{2,}$/,
  optEquals: /^(--\S+?)=(.*)/
}
```

ephox-mogran: I think the only real solution is to use yargs, because they've already solved this problem: https://github.com/yargs/yargs-parser/issues/145, but that's a bigger chunk of work.

modules/server/src/main/ts/bedrock/auto/Driver.ts
ephox-mogran: Removed the falsy check.

modules/server/src/test/ts/ClisTest.ts
spocke: Right that makes sense. I see now that the option description tells you about this requirement.

spocke (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 13.2.0 - 2022-05-10
+
+### Added
+- Added ability to use `--extraBrowserCapabilities` in bedrock-auto command.
```
lnewson, line 12:
```suggestion
- Additional webdriver browser capabilities can be passed via the `--extraBrowserCapabilities` argument in the bedrock-auto command.
```

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -67,6 +68,14 @@ const addArguments = (capabilities: Record<string, any>, name: string, args: str
   capabilities[name].args = currentArgs.concat(args);
 };
 
+const getExtraBrowserCapabilities = (settings: DriverSettings): string[] => {
+  if (settings.extraBrowserCapabilities.length > 0) {
```
lnewson, line 72: Should this be trimmed before comparing? As if it's just an empty string then you're not going to want a `[' ']` array to be returned.

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -188,6 +203,11 @@ export const create = async (settings: DriverSettings): Promise<Driver> => {
     // Wait for the driver to start up and then start the webdriver session
     await DriverLoader.startAndWaitForAlive(driverApi, port, webdriverTimeout);
     const webdriverOptions = getOptions(port, browserName, browserFamily, settings);
+
+    console.log(
```
lnewson, line 207: This seems like the wrong logging level here. This isn't something we really want logged every time as it's just noted needed. It should only be included when doing verbose logging (or I think it's called advanced logging in bedrock).

- [ ] Change this to only log when more detailed logging is enabled or remove it entirely

modules/server/src/main/ts/bedrock/cli/ClOptions.ts
```diff
@@ -264,6 +264,22 @@ export const useSandboxForHeadless: ClOption = {
   uncommon: true
 };
 
+/* In several situations, we need to pass through specific options to the
```
lnewson, line 267:
```suggestion
/*
 * In several situations, we need to pass through specific options to the 
```

lnewson (approved):

modules/server/src/main/ts/bedrock/auto/Driver.ts
ephox-mogran: From memory, advanced logging isn't really about log level ... it's about having a logging session that allows you to use special characters to move the cursor in the output and overwrite existing lines. I can put it through advanced logging, but I didn't realise we were using it for log levels. Do we do that anywhere else? 

I would strongly discourage not logging it altogether.

It's also just one to two lines. Is that really too much? Or does it look like way too much for Firefox Headless? Examples: 

```
e.g. chrome-headless
Browser capabilities: {"browserName":"chrome","goog:chromeOptions":{"args":["--start-maximized","--disable-extensions","--dig","--goose","--blah","--headless","--remote-debugging-port=9000"]}}

e.g. chrome
Browser capabilities: {"browserName":"chrome","goog:chromeOptions":{"args":["--start-maximized","--disable-extensions","--dig","--goose","--blah"]}

e.g. firefox
Browser capabilities: {"browserName":"firefox","moz:firefoxOptions":{"args":["--dig","--goose","--blah"]}}

e.g. firefox headless
Browser capabilities: {"browserName":"firefox","moz:firefoxOptions":{"args":["--dig","--goose","--blah","-headless","-start-debugger-server=9000"],"prefs":{"devtools.debugger.remote-enabled":true,"devtools.debugger.prompt-connection":false,"devtools.chrome.enabled":true}}}
```

I'm happy to do what you want, I'm just clarifying.

ephox-mogran: `advanced` logging is also the default, so I'm happy to change this to only log on advanced. But for clarification, someone is going to have to choose 'simple' for it not to appear.

- [ ] change to only log when mode is advanced.

modules/server/src/main/ts/bedrock/auto/Driver.ts
ephox-mogran: Good point.

- [ ] trim before comparing.

modules/server/src/main/ts/bedrock/auto/Driver.ts
lnewson: The problem is that by default we should only be providing information needed to know where to run the tests and where it's up to. You can't get anything meaningful from this logging except for debugging type information, so it makes little sense to show by default.

If `advanced` is already the default then we probably need to either add another mode or just remove this.

lnewson: Another thing worth noting is that the default capabilities aren't really meant to be known/visible, they are an internal implementation detail.

spocke: Maybe it should only log these if there are custom capabilities since then you might want to know what the full set is? And since that is more rare it would be logged less often only in the special cases like in a docker container or will that also be too verbose?

ephox-mogran: I think this is exactly the kind of thing that would have highlighted that Microsoft Edge Chromium doesn't have the flags, though. We can add a verbose flag if that's your preference, but not having any way of logging it without changing bedrock is going to make debugging a lot harder.

If you don't want it to be the default, then I'll need to add a verbose flag.

ephox-mogran: - [ ] add a verbose logging flag for extra details.

lnewson: Yeah having a verbose or debugging log level would definitely be the preferred option. Worth noting if we're doing that we may just want to enable verbose logging for webdriver.io as it includes all these details and much more about the webdriver calls.

lnewson (approved):

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -67,6 +69,15 @@ const addArguments = (capabilities: Record<string, any>, name: string, args: str
   capabilities[name].args = currentArgs.concat(args);
 };
 
+const getExtraBrowserCapabilities = (settings: DriverSettings): string[] => {
+  const extraCaps = settings.extraBrowserCapabilities.trim();
+  if (extraCaps.length > 0) {
```
spocke, line 74: No ternary I'm surprised. :wink:

spocke (approved):

modules/server/src/main/ts/bedrock/auto/Driver.ts
ephox-mogran: I thought it wasn't commonly used. I don't know what style to use any more. I'm just trying to fit in :)

lnewson: I was also surprised by this tbh 😂  I don't think we have a standard much anymore for this 😞

PR #100, opened 2022-05-16
TINY-8717: Removed the unused node-glob dependency

lnewson:
This removes the unused `node-glob` dependency which has deps on packages with security issues. It also hasn't been maintained in 6yrs. I've also upgraded the dev dependencies as well to resolve as many dependabot security alerts as I can.

### DIFF
--- a/modules/server/package.json
+++ b/modules/server/package.json
@@ -38,7 +38,6 @@
     "jquery": "^3.4.1",
     "mime-types": "^2.1.34",
     "mkdirp": "^1.0.0",
-    "node-glob": "^1.2.0",
     "portfinder": "^1.0.25",
     "recursive-readdir-sync": "^1.0.6",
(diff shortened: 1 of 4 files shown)


modules/server/package.json
```diff
@@ -63,9 +62,7 @@
     "@types/fresh": "^0.5.0",
     "@types/jquery": "^3.5.11",
     "@types/mime-types": "^2.1.1",
-    "@types/mkdirp": "^1.0.1",
-    "@types/webpack": "^5.28.0",
-    "@types/webpack-dev-server": "^4.5.0"
```
lnewson, line 68: I removed these as both webpack and webpack-dev-server now include types in their packages.

tiny-ben-tran (approved):

ltrouton (approved):

PR #101, opened 2022-06-08
TINY-8756: Added GitHub CODEOWNERS and PR template files

lnewson:


metricjs (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+Related Ticket:
```
spocke, line 1: I always wondered if this is really needed. Since the ticket ID is in the title or the PR. At least that is how I look these up when doing reviews.

spocke (approved):

.github/PULL_REQUEST_TEMPLATE.md
lnewson: Yeah I've thought the same, it just feels like a lot of duplication. I wonder if we should remove it while I'm making all these changes?

TheSpyder: It's useful for projects that autolink to JIRA, because PR titles don't get links.

But it does link in the commit messages so it's still going to be available _somewhere_.

lnewson: Ahh right, that was why we include it originally. I have no strong preferences here and if I don't need to go back and update ~40 repos I'm certainly not going to argue 😉

lnewson: Going to leave this as is for now.

PR #105, opened 2022-07-29
TINY-8944: Fixed the key effects leaving modifier keys active on Safari

lnewson:
Related Ticket: TINY-8944

Description of Changes:

Changed how the key effects logic works when using a combo key effect so that it uses the webdriver.io `keys` API instead to ensure modifier keys are released properly. Note that this can't be used at all times, as we need to use `elementSendKeys` when actually typing text otherwise it won't show.

I also fixed an issue I found when trying to debug whereby a `TypeError: Attempted to assign to readonly property` would get thrown instead on PhantomJS due to `stack` being a readonly property. So we just skip setting the stack in that case, even if it means the stack is no longer useful as it's only an issue on PhantomJS.

Pre-checks:
* [x] Changelog entry added
* [x] package.json version bumped (if first change of new major/minor)

Before merging:

### DIFF
--- a/modules/sample/src/test/ts/utils/Utils.ts
+++ b/modules/sample/src/test/ts/utils/Utils.ts
@@ -1,0 +1,42 @@
+import Promise from '@ephox/wrap-promise-polyfill';
+
+const post = (url: string, data: Record<string, any>): Promise<void> => {
+  return new Promise((onSuccess, onFailure) => {
+    const request = new XMLHttpRequest();
+    request.onreadystatechange = function () {
+      if (request.readyState === 4) {
+        if (request.status == 200) {
+          onSuccess();
+        } else {
+          onFailure();
+        }
+      }
+    };
+
+    request.onerror = function (err) {
+      debugger;
+      console.error(err);
+      onFailure();
+    };
+
+    request.open('POST', url);
+    request.setRequestHeader('Content-Type', 'application/json');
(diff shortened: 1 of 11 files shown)


HAFRMO (approved):

modules/sample/src/test/ts/utils/Utils.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (request.status == 200) {
+          onSuccess();
+        } else {
+          onFailure();
+        }
+      }
+    };
+
+    request.onerror = function (err) {
+      debugger;
```
ltrouton, line 17: Just wondering if we still want this debugger or should it be commented out?

ltrouton (approved):

modules/sample/src/test/ts/utils/Utils.ts
lnewson: Not sure to be honest... it was there in the `IFramePassTest.ts` where I copied this from, so not sure why it was added originally.

ephox-mogran (approved):

modules/sample/src/test/ts/utils/Utils.ts
lnewson: Going to leave it as is, as it's not going to harm anything in this case since it's not in the shipped code, it's just in the sample project which is only used by bedrock for its own tests.

PR #108, opened 2022-09-02
TINY-9097: Fixed lerna publish issues with new waluigi version

lnewson:
Related Ticket: TINY-9097

Description of Changes:

Lerna publish fails to run if there are uncommited changes and because we have to set npm config for the auth token this failed. So we now tell Git to ignore those changes before publishing.

Pre-checks:
* [x] ~Changelog entry added~
* [x] ~package.json version bumped (if first change of new major/minor)~

Before merging:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,7 +34,9 @@ tinyPods.node() {
   if (isReleaseBranch()) {
     stage("publish") {
       tinyNpm.withNpmPublishCredentials {
-        exec('yarn lerna publish from-package --yes --ignore @ephox/bedrock-sample')
+        // We need to tell git to ignore the changes to .npmrc when publishing
+        exec('git update-index --assume-unchanged .npmrc')
+        exec('yarn lerna publish from-package --yes --no-git-reset --ignore @ephox/bedrock-sample')
       }
     }
   }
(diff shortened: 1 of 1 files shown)


MitchCTiny (approved):

Jenkinsfile
```diff
@@ -5,36 +5,40 @@
...
 
-  if (isReleaseBranch()) {
-    stage("publish") {
-      tinyNpm.withNpmPublishCredentials {
-        exec('yarn lerna publish from-package --yes --ignore @ephox/bedrock-sample')
+    if (isReleaseBranch()) {
+      stage("publish") {
+        tinyNpm.withNpmPublishCredentials {
+          // We need to tell git to ignore the changes to .npmrc when publishing
+          exec('git update-index --assume-unchanged .npmrc')
```
spocke, line 39: Don't know enough about lerna. So in this case it willl git add all files and commit them on publish?

spocke (approved):

Jenkinsfile
lnewson: It doesn't actually end up doing anything with committing or pushing files and instead only tags things for the `from-package` because it's purpose is to just use what's already in git. Other variants of publish do though and the publish command as a whole will just try to check for uncommitted changes.

Anyways, the point of this addition is to make Git believe there are no actual changes to the npmrc file.

ltrouton (approved):

PR #123, opened 2023-06-01
TINY-9855: fix for watch mode

lorenzo-pomili:
Related Ticket: TINY-9855

Description of Changes:
the problem seems to be the 2 different version of `fork-ts-checker-webpack-plugin`

Pre-checks:
* [x] Changelog entry added
* [x] package.json version bumped (if first change of new major/minor)

Before merging:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- update `fork-ts-checker-webpack-plugin`. #TINY-9855
+
 ## 13.4.0 - 2023-05-17
 
 ### Improved
--- a/modules/server/package.json
+++ b/modules/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ephox/bedrock-server",
-  "version": "13.4.0",
+  "version": "13.4.1",
   "author": "Tiny Technologies Inc",
   "license": "Apache-2.0",
   "bin": {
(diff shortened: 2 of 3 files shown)


vpyshnenko (approved):

TheSpyder (changes requested):
Shouldn't `yarn.lock` be updated as well? Don't tell me you just edited the `package.json` 😉

lorenzo-pomili:
@TheSpyder I thought that I just had a last test to see if putting the dependencies back to the original makes me reproduce the bug again but I think it is not the case.

There is something odd with the `yarn install`, it seems that `--install.frozen-lockfile true` is not working as expected, you are allowed to launch `yarn install` with different dependencies versions in sub-directories even if you have `--install.frozen-lockfile true` and the dependencies are downloaded at the correct version, but the `yarn.lock` file is not updated.

I launched `yarn install` without `--install.frozen-lockfile true` to update the `yarn.lock` (of course it has updated a lot of other dependencies too), I'll check to see if there is a way to make it works also in the sub-directory (I already try to add the `.yarnrc` also to the sub-directory)

TheSpyder (approved):
To update `yarn.lock` without turning off frozen lockfile you just need to use `yarn add` or `yarn upgrade-interactive` instead of editing the `package.json` by hand.

I'm not sure why yarn let itself run with such a huge disparity in version numbers. Normally it fails saying the lockfile needs to be updated.

lorenzo-pomili:
@TheSpyder I think it is related to the fact that it is not at the first level :/

PR #131, opened 2023-08-07
TINY-10006: Add `--remote` flag for remote testing

Skylite73:
Related Ticket: TINY-10006

Description of Changes:
* Add `--remote` flag to support remote testing using third parties (lambdatest and Aws device farm)

Pre-checks:
* [x] Changelog entry added

Before merging:

### DIFF
--- a/modules/server/src/main/ts/BedrockAuto.ts
+++ b/modules/server/src/main/ts/BedrockAuto.ts
@@ -34,7 +34,9 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
       useSandboxForHeadless: settings.useSandboxForHeadless,
       extraBrowserCapabilities: settings.extraBrowserCapabilities,
       verbose: settings.verbose,
-      wipeBrowserCache: settings.wipeBrowserCache
+      wipeBrowserCache: settings.wipeBrowserCache,
+      webdriverPort: settings.webdriverPort,
+      useSelenium: settings.useSelenium
     });
 
     const webdriver = driver.webdriver;
--- a/modules/server/src/main/ts/BedrockAuto.ts
+++ b/modules/server/src/main/ts/BedrockAuto.ts
@@ -19,33 +19,36 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
   console.log('bedrock-auto ' + Version.get() + ' starting...');
 
   const settings = SettingsResolver.resolveAndLog(bedrockAutoSettings);
-  console.log('Bedrock settings:', settings);
   const master = DriverMaster.create();
   const browserName = settings.browser.replace('-headless', '');
   const isPhantom = browserName === 'phantomjs';
   const isHeadless = settings.browser.endsWith('-headless') || isPhantom;
   const basePage = 'src/resources/html/' + (isPhantom ? 'bedrock-phantom.html' : 'bedrock.html');
+  // Remote settings
   const remoteWebdriver = settings.remote;
+  const sishDomain = settings.sishDomain;
+  const username = settings.username ?? process.env.LT_USERNAME;
+  const accesskey = settings.accesskey ?? process.env.LT_ACCESS_KEY;
 
   const routes = RunnerRoutes.generate('auto', settings.projectdir, settings.basedir, settings.config, settings.bundler, settings.testfiles, settings.chunk, settings.retries, settings.singleTimeout, settings.stopOnFailure, basePage, settings.coverage, settings.polyfills);
 
   routes.then(async (runner) => {
-     const shutdownServices: (() => Promise<any>)[] = [];
+    const shutdownServices: (() => Promise<any>)[] = [];
 
     // LambdaTest Tunnel must know dev server port, but tunnel must be created before dev server.
     const servicePort = await portfinder.getPortPromise({
       port: 8000,
       stopPort: 20000
     });
-    let location = 'http://localhost:' + servicePort;
-    if (remoteWebdriver === 'lambdatest') {
-      const tunnel = await Tunnel.create(remoteWebdriver, servicePort);
-      shutdownServices.push(tunnel.shutdown);
-    } else if (remoteWebdriver === 'aws') {
-      const tunnel = await Tunnel.create(remoteWebdriver, servicePort);
-      location = tunnel.url.href;
-      shutdownServices.push(tunnel.shutdown);
-    }
+
+    const tunnelCredentials = {
+      user: username,
+      key: accesskey
+    };
+
+    const tunnel = await Tunnel.prepareConnection(servicePort, remoteWebdriver, sishDomain, tunnelCredentials);
+    shutdownServices.push(tunnel.shutdown);
+    const location = tunnel.url.href;
 
     console.log('Creating webdriver...');
     if (remoteWebdriver == 'aws') {
@@ -62,7 +65,11 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
       wipeBrowserCache: settings.wipeBrowserCache,
       remoteWebdriver,
       webdriverPort: settings.webdriverPort,
-      useSelenium: settings.useSelenium
+      useSelenium: settings.useSelenium,
+      username,
+      accesskey,
+      devicefarmRegion: settings.devicefarmRegion,
+      deviceFarmArn: settings.devicefarmArn
     });
 
     const webdriver = driver.webdriver;
@@ -75,6 +82,8 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
       port: servicePort
     });
     shutdownServices.push(service.shutdown, driver.shutdown);
+
+    const shutdown = () => Promise.allSettled(shutdownServices.map((shutdown_fn) => shutdown_fn()));
 
     try {
       if (!isHeadless) {
@@ -95,13 +104,8 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
         return Reporter.writePollExit(settings, data);
       });
 
-      // Combine all shutdowns into single function call.)
-      const shutdown = () => Promise.allSettled(shutdownServices.map((shutdown_fn) => shutdown_fn()));
-
       return Lifecycle.done(result, webdriver, shutdown, settings.gruntDone, settings.delayExit);
     } catch (e) {
-      // Combine all shutdowns into single function call.
-      const shutdown = () => Promise.allSettled(shutdownServices.map((shutdown_fn) => shutdown_fn()));
       return Lifecycle.error(e, webdriver, shutdown, settings.gruntDone, settings.delayExit);
     }
   }).catch((err) => {
--- a/modules/server/src/main/ts/BedrockAuto.ts
+++ b/modules/server/src/main/ts/BedrockAuto.ts
@@ -6,26 +6,51 @@ import * as RunnerRoutes from './bedrock/server/RunnerRoutes';
 import * as Reporter from './bedrock/core/Reporter';
 import * as DriverMaster from './bedrock/server/DriverMaster';
 import * as Driver from './bedrock/auto/Driver';
+import * as Tunnel from './bedrock/auto/Tunnel';
 import * as Lifecycle from './bedrock/core/Lifecycle';
 import { BedrockAutoSettings } from './bedrock/core/Settings';
 import { ExitCodes } from './bedrock/util/ExitCodes';
 import * as ConsoleReporter from './bedrock/core/ConsoleReporter';
 import * as SettingsResolver from './bedrock/core/SettingsResolver';
+import * as portfinder from 'portfinder';
+import { format } from 'node:util';
 
 export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
   console.log('bedrock-auto ' + Version.get() + ' starting...');
 
   const settings = SettingsResolver.resolveAndLog(bedrockAutoSettings);
+  console.log('Bedrock settings:', settings);
   const master = DriverMaster.create();
   const browserName = settings.browser.replace('-headless', '');
   const isPhantom = browserName === 'phantomjs';
   const isHeadless = settings.browser.endsWith('-headless') || isPhantom;
   const basePage = 'src/resources/html/' + (isPhantom ? 'bedrock-phantom.html' : 'bedrock.html');
+  const remoteWebdriver = settings.remote;
 
   const routes = RunnerRoutes.generate('auto', settings.projectdir, settings.basedir, settings.config, settings.bundler, settings.testfiles, settings.chunk, settings.retries, settings.singleTimeout, settings.stopOnFailure, basePage, settings.coverage, settings.polyfills);
 
   routes.then(async (runner) => {
+     const shutdownServices: (() => Promise<any>)[] = [];
 
+    // LambdaTest Tunnel must know dev server port, but tunnel must be created before dev server.
+    const servicePort = await portfinder.getPortPromise({
+      port: 8000,
+      stopPort: 20000
+    });
+    let location = 'http://localhost:' + servicePort;
+    if (remoteWebdriver === 'lambdatest') {
+      const tunnel = await Tunnel.create(remoteWebdriver, servicePort);
+      shutdownServices.push(tunnel.shutdown);
+    } else if (remoteWebdriver === 'aws') {
+      const tunnel = await Tunnel.create(remoteWebdriver, servicePort);
+      location = tunnel.url.href;
+      shutdownServices.push(tunnel.shutdown);
+    }
+
+    console.log('Creating webdriver...');
+    if (remoteWebdriver == 'aws') {
+        console.log('INFO: Webdriver creation waits for device farm session to activate. Takes 30-45s.');
+    }
     const driver = await Driver.create({
       browser: browserName,
       basedir: settings.basedir,
@@ -35,6 +60,7 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
       extraBrowserCapabilities: settings.extraBrowserCapabilities,
       verbose: settings.verbose,
       wipeBrowserCache: settings.wipeBrowserCache,
+      remoteWebdriver,
       webdriverPort: settings.webdriverPort,
       useSelenium: settings.useSelenium
     });
@@ -45,17 +71,18 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
       driver: Attempt.passed(webdriver),
       master,
       runner,
-      stickyFirstSession: true
+      stickyFirstSession: true,
+      port: servicePort
     });
-
-    const shutdown = () => Promise.all([ service.shutdown(), driver.shutdown() ]);
+    shutdownServices.push(service.shutdown, driver.shutdown);
 
     try {
       if (!isHeadless) {
-        console.log('bedrock-auto ' + Version.get() + ' available at: http://localhost:' + service.port);
+        console.log('bedrock-auto ' + Version.get() + ' available at: ' + location);
       }
 
-      await webdriver.url('http://localhost:' + service.port);
+      console.log('Loading initial page...');
+      await webdriver.url(location);
       console.log(isPhantom ? '\nPhantom tests loading ...\n' : '\nInitial page has loaded ...\n');
       service.markLoaded();
       service.enableHud();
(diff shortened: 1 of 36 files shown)


TheSpyder:
At least set a reasonable title and description please 😛

modules/server/src/main/ts/bedrock/auto/Tunnel.ts
```diff
@@ -0,0 +1,116 @@
...
+      reject('Tunnel did not yield a URL');
+      return;
+    }
+
+    const wordStream = output.pipe(split2());
+    wordStream.on('data', (_) => {
+      // console.log('Tunnel says:', data.toString());
+      // Stop listening. Even though it resolves instantly, this code block would otherwise
+      // still be run on every data event.
+      // wordStream.removeAllListeners('data');
```
ltrouton, line 32: Not sure if this comment needs to be cleaned up at all

modules/server/src/main/ts/bedrock/auto/Tunnel.ts
```diff
@@ -0,0 +1,116 @@
...
+import { Tunnel as LambdaTunnel } from '@lambdatest/node-tunnel';
+
+export interface Tunnel {
+  url: URL;
+  shutdown: () => Promise<void>;
+}
+
+const loadSSH = (subdomain: string, port: number | string): ExecUtils.ChildAPI => {
+  const exec = 'ssh';
+  const args = ['-R', subdomain + ':80:localhost:' + port, 'sish.osu.tiny.work'];
```
ltrouton, line 14: Just wondering if `'sish.osu.tiny.work'` should be configurable via the the CLI and not baked in. Bedrock is technically a public repo so we have to be mindful of that and assume I we want to keep everything agnostic

modules/server/src/main/ts/BedrockAuto.ts
```diff
@@ -6,26 +6,51 @@ import * as RunnerRoutes from './bedrock/server/RunnerRoutes';
...
 import * as ConsoleReporter from './bedrock/core/ConsoleReporter';
 import * as SettingsResolver from './bedrock/core/SettingsResolver';
+import * as portfinder from 'portfinder';
+import { format } from 'node:util';
 
 export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
   console.log('bedrock-auto ' + Version.get() + ' starting...');
 
   const settings = SettingsResolver.resolveAndLog(bedrockAutoSettings);
+  console.log('Bedrock settings:', settings);
```
ltrouton, line 22: Do we want this logged?
```suggestion
```

modules/server/src/main/ts/BedrockAuto.ts
```diff
@@ -6,26 +6,51 @@ import * as RunnerRoutes from './bedrock/server/RunnerRoutes';
...
   const browserName = settings.browser.replace('-headless', '');
   const isPhantom = browserName === 'phantomjs';
   const isHeadless = settings.browser.endsWith('-headless') || isPhantom;
   const basePage = 'src/resources/html/' + (isPhantom ? 'bedrock-phantom.html' : 'bedrock.html');
+  const remoteWebdriver = settings.remote;
 
   const routes = RunnerRoutes.generate('auto', settings.projectdir, settings.basedir, settings.config, settings.bundler, settings.testfiles, settings.chunk, settings.retries, settings.singleTimeout, settings.stopOnFailure, basePage, settings.coverage, settings.polyfills);
 
   routes.then(async (runner) => {
+     const shutdownServices: (() => Promise<any>)[] = [];
```
spocke, line 33: Weird indentation.

```suggestion
    const shutdownServices: (() => Promise<any>)[] = [];
```

modules/server/src/main/ts/BedrockAuto.ts
```diff
@@ -68,12 +95,18 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
         return Reporter.writePollExit(settings, data);
       });
 
+      // Combine all shutdowns into single function call.)
+      const shutdown = () => Promise.allSettled(shutdownServices.map((shutdown_fn) => shutdown_fn()));
```
spocke, line 99: Could this be defined once seems to be the same function. Also is the comment overkill, it is kind of obvious what the function does.

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -80,16 +85,19 @@ const getExtraBrowserCapabilities = (settings: DriverSettings): string[] => {
...
-    port,
-    logLevel: 'silent' as const,
+  const options: WebdriverIO.RemoteOptions = {
+    logLevel: 'warn' as const,
     capabilities: {
       browserName
     }
   };
 
+  if (!settings.remoteWebdriver) {
```
spocke, line 95: Can we move this as a else statement for the if on remoteWebDriver below it seems that the settings is not really used until later. This makes the code easier to read since the branching on this is in one place.

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -130,10 +139,33 @@ const getOptions = (port: number, browserName: string, settings: DriverSettings,
     }
   }
 
+  // Remote webdriver settings
```
spocke, line 142: This is kind of obvious. So the comment is not really needed.

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -130,10 +139,33 @@ const getOptions = (port: number, browserName: string, settings: DriverSettings,
...
+  if (settings.remoteWebdriver) {
+
+    caps.browserVersion = 'latest';
+
+    if (browserName == 'firefox') {
```
spocke, line 147: These things are not obvious why are they needed, what do they do?

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -213,6 +245,74 @@ const getDriverSpec = (settings: DriverSettings, browserName: string): DriverLoa
...
+  } else {
+    await driver.maximizeWindow();
+  }
+  
+  return Promise.resolve();
+};
+
+const getFarmUrl = async (): Promise<URL> => {
+  console.log('Creating DeviceFarmClient...');
+  const client = new DeviceFarmClient({region: 'us-west-2'});
```
spocke, line 270: Should the region here be configurable as a setting or be pulled from env?

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -213,6 +245,74 @@ const getDriverSpec = (settings: DriverSettings, browserName: string): DriverLoa
...
+  const input = {
+    projectArn: 'arn:aws:devicefarm:us-west-2:103651136441:testgrid-project:79ff2b40-fe26-440f-9539-53163c25442e',
```
spocke, line 272: This should be configurable or pulled from env right? We could default it if it becomes annoying to specify.

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -213,6 +245,74 @@ const getDriverSpec = (settings: DriverSettings, browserName: string): DriverLoa
...
+    }
+    const url = await getFarmUrl();
+
+    const options = deepmerge(defaultSettings, {
+      hostname: url.host,
+      path: url.pathname,
+      protocol: 'https',
+      port: 443,
+      capabilities: {
+        'aws:maxDurationSecs': 2400,
```
spocke, line 296: What does this do? Kill the test if it runs longer than this?

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -213,6 +245,74 @@ const getDriverSpec = (settings: DriverSettings, browserName: string): DriverLoa
...
+      },
+    });
+
+    console.log('Starting Device Farm session with options:', JSON.stringify(options, null, 2));
```
spocke, line 300: Could the node formatter that was pulled in be used here or is that not the same thing as a stringify?

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -221,61 +321,66 @@ const getDriverSpec = (settings: DriverSettings, browserName: string): DriverLoa
...
-
-    const webdriverOptions = getOptions(port, browserName, settings, debuggingPort);
-
-    if (settings.verbose) {
-      console.log(
-        `Browser capabilities: ${JSON.stringify(webdriverOptions.capabilities)}`
-      );
-    }
-
+  // Find an open port to start the driver on
```
spocke, line 327:
```suggestion

  // Find an open port to start the driver on
```

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -221,61 +321,66 @@ const getDriverSpec = (settings: DriverSettings, browserName: string): DriverLoa
...
+  const port = await getPort(settings.webdriverPort, 4444);
+  const debuggingPort = settings.headless ? await getPort(settings.debuggingPort, 9000) : 9000;
+  const webdriverOptions = getOptions(port, browserName, settings, debuggingPort);
+  console.log('Webdriver options:', webdriverOptions);
```
spocke, line 331: Does these options need to be formatted or does it print it in a nice way?

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
```diff
@@ -31,64 +21,34 @@ const browserExecutables: Record<string, string[]> = {
...
-
 const loadPhantomJs = (settings: DriverSettings) => {
-  const api = execLoader('phantomjs', [ '--remote-debugger-port=' + settings.debuggingPort ]);
+  const api = ExecUtils.execLoader('phantomjs', [ '--remote-debugger-port=' + settings.debuggingPort ]);
 
   // Patch the start function to remap the arguments
   const origStart = api.start;
   api.start = (args = []) => {
-    const patchedArgs = args.map((arg) => {
+    const patchedArgs = args.map((arg: any) => {
```
spocke, line 30: I wonder if we should circle back on this one and fix the types. It seems that all invocations of this function is either without arguments or with a first argument with an array of strings. I'm always a bit uneasy when the `any` word goes in but this seems to have been like this for a long time.

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
```diff
@@ -102,78 +62,29 @@ export const loadDriver = (browserName: string, settings: DriverSettings): Drive
...
 
   const execNames = browserExecutables[browserName] || [];
-  const execPath = findExecutable(execNames);
+  const execPath = ExecUtils.findExecutable(execNames);
   if (execPath !== null) {
-    return execLoader(execPath);
+    return ExecUtils.execLoader(execPath);
   } else {
-    throw new Error('Unable to find a suitable driver for ' + browserName);
+    throw new Error('Local1: Unable to find a suitable driver for ' + browserName);
```
spocke, line 69: `Local1` ?  What is that?

modules/server/src/test/ts/ClisTest.ts
```diff
@@ -60,8 +60,9 @@ describe('Clis.forAuto', () => {
       useSelenium: false,
       extraBrowserCapabilities: '',
       skipResetMousePosition: false,
-      webdriverPort: 4444,
-      wipeBrowserCache: false
+      wipeBrowserCache: false,
+      remote: '',
```
spocke, line 64: Should there be a test for the remote options?

spocke (approved):

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
```diff
@@ -102,78 +62,29 @@ export const loadDriver = (browserName: string, settings: DriverSettings): Drive
...
+export const startAndWaitForAlive = async (driverSpec: DriverSpec, port: number, timeout = 30000): Promise<void> => {
+  const api = driverSpec.driverApi;
+  
+  // Start the driver
+  let args;
+  if (api.isNpm) {
+    args = {port};
+  } else {
+    args = ['--port=' + port];
+  }
```
MitchC1999, line 82:
```suggestion
  const args = api.isNpm ? { port } : [ '--port=' + port ];
```
If this is not to be changed. Might still need typing.

MitchC1999 (approved):
Some variables which could require updating in the future may be better off in a globals file, or as default settings. Looks like this'll need thorough testing on our part. Nice work on all of this.

modules/server/src/main/ts/bedrock/auto/Tunnel.ts
jscasca: Changed to accept everything as args or read them from the env

modules/server/src/main/ts/BedrockAuto.ts
jscasca: Actually removed this log

modules/server/src/main/ts/BedrockAuto.ts
jscasca: Weird formats were addressed

modules/server/src/main/ts/BedrockAuto.ts
jscasca: Removed duplication

modules/server/src/main/ts/bedrock/auto/Driver.ts
jscasca: Changed how opts for remote are specified. Driver opts are merged before returning. Ideally, would like to have a more functional way of adding arguments to opts and capabilities but seemed like a bigger change at the time

modules/server/src/main/ts/bedrock/auto/Driver.ts
jscasca: Removed this comment

modules/server/src/main/ts/bedrock/auto/Driver.ts
jscasca: They get printed nicely in node. Only down to one level which should be good enough for shallow issues

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
jscasca: Was used for local debugging. Has been removed

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
jscasca: These particular types are tricky because of the driver differences between loading via npm and having executable drivers and typescript not behaving nicely with rest parameters. At least at this point we can assume `string`s but typing the rest parameters did not go as expected, unfortunately

modules/server/src/main/ts/bedrock/auto/DriverLoader.ts
jscasca: Types for that `driverApi.start` have not been an easy task due to it receiving rest parameters and typescript being weird about it. But at least it got inlined as suggested

modules/server/src/main/ts/bedrock/auto/Driver.ts
TheSpyder: Yes bedrock arguments are weird and I generally hate how fragile they are

modules/server/src/main/ts/bedrock/cli/ClOptions.ts
```diff
@@ -43,7 +43,8 @@ export const browser: ClOption = {
     'ie',
     'firefox',
     'firefox-headless',
-    'MicrosoftEdge',
+    'MicrosoftEdge', // ? I doubt this should be here.
```
TheSpyder, line 46: Not only is it necessary, if you say `edge` it aliases back to this
https://github.com/tinymce/bedrock/blob/3912e4e00d04809596a38044947d910dc012787d/modules/server/src/main/ts/bedrock/auto/Driver.ts#L29

It's weird but updating all the myriad places that refer to this as `MicrosoftEdge` is too painful at the moment. So we aliased it to load the edge-chromium driver even though that's not what this normally refers to.

We have two options:
* Leave two edge references in here, which is weird and will probably break if someone tries to use `edge` for local testing
* Add an alias in the remote driver so `MicrosoftEdge` becomes `edge`, which is perhaps even more weird
```suggestion
    'MicrosoftEdge',
```

modules/server/src/test/ts/ClisTest.ts
TheSpyder: yes, but there's not enough dev time left in the schedule

modules/server/src/main/ts/bedrock/auto/Driver.ts
jscasca: Space added

TheSpyder: The inline diff makes it look like an extra line is required, but look side-by-side and it's fine

modules/server/src/main/ts/bedrock/util/ExecUtils.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+export const execLoader = (exec: string, args: string[] = []): ChildAPI => {
+  const api = childAPIWrapper(crossSpawn, false, exec, args);
+  return api;
+};
+
+// Npm drivers (e.g. geckodriver) are called differently to binaries. This is to conform them
+// into a single compatible type. I don't think npm drivers were used until WDIOv8, which has them
+// installed by default as dependencies.
```
TheSpyder, line 22: That's correct, and it's actually quite bad for us. Having drivers as direct dependencies seems like a nice way to reduce the load on the ops team, until you realise that means browser upgrades require _every project to upgrade bedrock_, rather than a single server upgrade - the server upgrades have been painful, but upgrading 50 repositories would be more painful.

I suppose the premium monorepo gets to eat this lunch, as once it is ready there will only be two place to upgrade bedrock, but that's still painful for every active branch when a browser upgrade happens.

Even so, I figure one day we'll end up removing the non-npm driver code.

TheSpyder (approved):
I've probably said this before, but I feel even more bad now that we have lost so many bedrock experts you needed to figure this out on your own.

The original way this was implemented - as a separate `bedrock-remote` command instead of `bedrock-auto` - was way easier to manage than wedging it in as an option has become. But that was removed, and the code has moved on since then.

It's done now, good work, we'll get this through and start figuring out what else needs to be changed. Once it's stable I plan to remove all the spammy console logs but I figure they're very helpful for debugging issues at this stage.

modules/server/src/main/ts/bedrock/cli/ClOptions.ts
jscasca: Happy to keep it as s for this release and note it down

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -213,6 +233,26 @@ const getDriverSpec = (settings: DriverSettings, browserName: string): DriverLoa
   };
 };
 
+const driverSetup = async (driver: WebdriverIO.Browser, settings: DriverSettings, debuggingPort: number): Promise<void> => {
+  // Browsers have a habit of reporting via the webdriver that they're ready before they are (particularly FireFox).
+
+  // setTimeout is a temporary solution, VAN-66 has been logged to investigate properly
```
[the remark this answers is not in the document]
jscasca: I'll just leave this one for the paper trail as it doesn't change the behaviour of the current solution, we just moved code around this.

modules/server/src/main/ts/bedrock/auto/Driver.ts
jscasca: Variables are taken as args or read from env as suggested

modules/server/src/main/ts/bedrock/auto/Driver.ts
jscasca: Yes, that's the maximum allowed value before a session is forcibly closed. Comment has been added to the code

modules/server/src/main/ts/bedrock/auto/Driver.ts
jscasca: This logging has been removed

modules/server/src/main/ts/bedrock/auto/Driver.ts
jscasca: Comments have been added to the code. In this case, we added better logging for firefox for readability purposes

PR #134, opened 2023-10-31
TINY-10275: Removed clipboard effects

spocke:
Related Ticket: TINY-10275

Description of Changes:
* Removed the `/clipboard` effects
* Removed a test for that
* Removed the `wink.exe`
* Fixed some mistakes around formatting around the last release changelog item.

- [ ] Bump major before release

Pre-checks:
* [x] Changelog entry added

Before merging:


vpyshnenko (approved):

TheSpyder (approved):
I'm borderline on this one. We might still need it to test TinyMCE LTS.

At the very least please don't merge it until all tests that use it have been replaced, in case we need to do an emergency bedrock patch for something.

- [ ] wait until the other PRs have merged

spocke:
> I'm borderline on this one. We might still need it to test TinyMCE LTS.

Maybe if we have the infra to run it technically this is only for sanity testing we could make that a manual step for LTS unless we don't have one. It's also only testing this on windows so nothing is tested on other browsers. We could also use older bedrock before this change for LTS.

> At the very least please don't merge it until all tests that use it have been replaced, in case we need to do an emergency bedrock patch for something.

Yes, I wanted to get this up for review before I fixed the tests. But I can't fix the tests until the agar change is merged and published as a alpha so right now kind of stuck.

spocke:
Maybe we park this for a while still think we should get rid of code that we no longer use but there is no rush for it. But if we shutdown the current VMs then this effect will no longer work unless you run the test in local windows VM.

PR #139, opened 2024-01-29
TINY-10604: LambdaTest improvements

jscasca:
- Pass tunnel name for webdriver
- Use LT project naming
- Mark LT test results

Related Ticket: TINY-10604

Description of Changes:
* Placeholder text

Pre-checks:
* [x] Changelog entry added
* [-] package.json versions have not been changed (done by Lerna on release)
* [-] Tests have been added (if applicable)

Before merging:
* [-] Ensure internal dependencies are on appropriate versions

### DIFF
--- a/modules/server/src/main/ts/BedrockAuto.ts
+++ b/modules/server/src/main/ts/BedrockAuto.ts
@@ -32,8 +32,8 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
 
   const routes = RunnerRoutes.generate('auto', settings.projectdir, settings.basedir, settings.config, settings.bundler, settings.testfiles, settings.chunk, settings.retries, settings.singleTimeout, settings.stopOnFailure, basePage, settings.coverage, settings.polyfills);
 
-  const shutdownServices: (() => Promise<any>)[] = [];
-  const shutdown = (services: (() => Promise<any>)[]) => () => Promise.allSettled(services.map((fn) => fn()));
+  const shutdownServices: (() => Promise<void>)[] = [];
+  const shutdown = (services: (() => Promise<void>)[]) => () => Promise.allSettled(services.map((fn) => fn()));
 
   routes.then(async (runner) => {
 
--- a/modules/server/src/main/ts/BedrockAuto.ts
+++ b/modules/server/src/main/ts/BedrockAuto.ts
@@ -32,8 +32,10 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
 
   const routes = RunnerRoutes.generate('auto', settings.projectdir, settings.basedir, settings.config, settings.bundler, settings.testfiles, settings.chunk, settings.retries, settings.singleTimeout, settings.stopOnFailure, basePage, settings.coverage, settings.polyfills);
 
+  const shutdownServices: (() => Promise<any>)[] = [];
+  const shutdown = (services: (() => Promise<any>)[]) => () => Promise.allSettled(services.map((fn) => fn()));
+
   routes.then(async (runner) => {
-    const shutdownServices: (() => Promise<any>)[] = [];
 
     // LambdaTest Tunnel must know dev server port, but tunnel must be created before dev server.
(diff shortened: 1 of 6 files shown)


TheSpyder (approved):

modules/server/src/main/ts/BedrockAuto.ts
```diff
@@ -32,8 +32,10 @@ export const go = (bedrockAutoSettings: BedrockAutoSettings): void => {
 
   const routes = RunnerRoutes.generate('auto', settings.projectdir, settings.basedir, settings.config, settings.bundler, settings.testfiles, settings.chunk, settings.retries, settings.singleTimeout, settings.stopOnFailure, basePage, settings.coverage, settings.polyfills);
 
+  const shutdownServices: (() => Promise<any>)[] = [];
```
spocke, line 35: Could these use `T` or are the services all random types?

modules/server/src/main/ts/bedrock/core/Lifecycle.ts
```diff
@@ -12,6 +12,10 @@ const exitDelay = (driver: Browser, delayExiting: boolean) => {
   return delayExiting ? driver.pause(17 * 60 * 1000) : Promise.resolve();
 };
 
+const markLambdaTest = async (driver: Browser, status: 'failed' | 'passed') => {
+  await driver.executeScript('if (window.lambda_status) { lambda_status=' + status + ' }', []);
```
spocke, line 16: Could we link this to the docs and also add some comment why it's needed. I can't seem to find `lambda_status` all I see is this that uses `lambda-status` didn't look very hard though.

modules/server/src/main/ts/bedrock/auto/RemoteDriver.ts
```diff
@@ -70,14 +70,29 @@ export const getApi = async (settings: DriverSettings, browser: string, opts: We
...
-      shutdown: () => Promise.resolve()
+      shutdown: () => driver.deleteSession()
     };
   }
   return Promise.reject('Unrecognized remote provider: [' + remoteWebdriver + ']');
 };
 
 const addDriverSpecificOpts = (opts: WebdriverIO.RemoteOptions, settings: DriverSettings): WebdriverIO.RemoteOptions => {
   if (settings.remoteWebdriver === 'lambdatest') {
+    // For naming in LT we use PROJECT_BUILD[_NAME] or BUILD
```
spocke, line 81: Do we enforce this format in the settings so that we don't pass in some format that lambda test doesn't support? And is that format documented somewhere can we point to it.

spocke (approved):

modules/server/src/main/ts/bedrock/auto/RemoteDriver.ts
jscasca: LT allows user to specify a `build`, `project` and test `name` as part of the options map. There is no specific format. I didn't want to keep adding Client Options to bedrock so my idea was to get that from the name. If no name is passed it defaults to `bedrock-auto` for the name of the build. We can either add the extra effort of accepting client options but it only will be used for LT or we can identify one string pattern to pass via the name

modules/server/src/main/ts/bedrock/core/Lifecycle.ts
spocke: Ok could you write some comments in the code here or refer to more information in the ticket. Just want to avoid the next dev being confused about this does in N months from now and then having to search though forums etc to find the thing again.

modules/server/src/main/ts/BedrockAuto.ts
jscasca: Changed to correct types: `Promise<void>`

modules/server/src/main/ts/bedrock/core/Lifecycle.ts
jscasca: Linked documentation in ticket and added relevant information. Added comments to the code and the ticket reference

PR #141, opened 2024-02-29
TINY-10688: Support using `pnpm-workspace.yaml`

danoaky-tiny:
Related Ticket: TINY-10688

Description of Changes:
If the `workspaces` prop inside of the project's `package.json` is unused, check if there's a `pnpm-workspace.yaml` file then use `pnpm list -r --only-projects --json` to find workspaces instead.

Pre-checks:
* [x] Changelog entry added
* [X] ~Tests have been added (if applicable)~

Before merging:

### DIFF
--- a/modules/server/src/main/ts/bedrock/server/RunnerRoutes.ts
+++ b/modules/server/src/main/ts/bedrock/server/RunnerRoutes.ts
@@ -1,5 +1,6 @@
 import * as path from 'path';
 import * as fs from 'fs';
+import * as childProcess from 'child_process';
 import * as glob from 'glob';
 import * as Routes from './Routes';
 import * as Compiler from '../compiler/Compiler';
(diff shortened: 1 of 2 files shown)


ltrouton (approved):

modules/server/src/main/ts/bedrock/server/RunnerRoutes.ts
```diff
@@ -41,10 +47,32 @@ export const generate = async (mode: string, projectdir: string, basedir: string
...
-  const workspaceRoots = (
+  const findPnpmWorkspaces = async (): Promise<WorkspaceRoot[]> => {
+    if (!fs.existsSync(path.join(projectdir, 'pnpm-workspace.yaml'))) {
+      return [];
+    }
+
+    return new Promise<WorkspaceRoot[]>((resolve, reject) =>
+      childProcess.exec('pnpm list -r --only-projects --json', (err, stdout, stderr) => {
+        if (err) reject(err);
+        else if (stderr) reject(stderr);
```
TheSpyder, line 58: This doesn't necessarily mean the command failed 🤔 

`stderr` should be printed, though, regardless of error state.

TheSpyder (approved):
Note bedrock doesn't have an automated release process, so you'll probably need help from an admin to publish once we're ready to do a new release

modules/server/src/main/ts/bedrock/server/RunnerRoutes.ts
danoaky-tiny: `stderr` is just error logged now. It only rejects if there's an `err`

PR #142, opened 2024-03-05
TINY-10708: Suppress bedrock logs in remote testing

danoaky-tiny:
Related Ticket: TINY-10708

Description of Changes:
- Reduce logs when on CI to 10% intervals.
- Cursor is no longer moved (or at least tries to) when on CI, so logs are just normally printed.

Pre-checks:
* [x] Changelog entry added
* [X] package.json versions have not been changed (done by Lerna on release)
* [X] ~Tests have been added (if applicable)~

Before merging:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Improved
+
+- Reduced progress logs when on CI to 10% intervals #TINY10708
+
 ## 14.1.2 - 2024-01-31
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 
-- Reduced progress logs when on CI to 10% intervals #TINY10708
+- Reduced progress logs when on CI to 10% intervals #TINY-10708
 
 ## 14.1.2 - 2024-01-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Improved
 
 - Reduced progress logs when on CI to 10% intervals #TINY-10708
+- Now supports using `pnpm-workspace.yaml` to fetch workspaces #TINY-10688
 
 ## 14.1.2 - 2024-01-31
 
--- a/modules/server/src/main/ts/bedrock/cli/Hud.ts
+++ b/modules/server/src/main/ts/bedrock/cli/Hud.ts
@@ -38,7 +38,7 @@ export const create = (testfiles: string[], loglevel: 'simple' | 'advanced'): Hu
     return Promise.resolve();
   };
 
-  const advUpdate = async (data: ResultData) => {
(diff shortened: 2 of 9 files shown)


TheSpyder (changes requested):
Looks pretty good. Rather than using remote as the flag, can you please use a TTY check and maybe look for the CI environment variable?

I want this change for all CI, including headless tests.

modules/server/src/main/ts/bedrock/server/Controller.ts
```diff
@@ -107,9 +108,16 @@ export const create = (stickyFirstSession: boolean, singleTimeout: number, overa
...
+    if (stickyFirstSession && (timeoutError || session.id !== stickyId)) return false;
+    if (!Env.IS_CI || session.done) return true;
+    // Only update the HUD at 10% intervals on remote:
+    return session.results.length % Math.round(session.totalTests * 0.1) === 0;
+  };
+
   const updateHud = (session: TestSession) => {
-    if (!outputToHud) return;
-    if (stickyFirstSession && (timeoutError || session.id !== stickyId)) return;
+    if (!shouldUpdateHud(session)) return;
```
TheSpyder, line 120: I'm torn about whether to keep this on the server or just reduce the frequency of traffic from the client.

Lower traffic should speed up the connection on slow tunnels, but might risk incorrect HUD output when a test times out.

TheSpyder (approved):
Much better 👍

modules/server/src/main/ts/bedrock/server/Controller.ts
danoaky-tiny: Maybe it should always log tests that didn't pass as well?
```typescript
if (!Env.IS_CI || session.done || !session.results.at(-1)?.passed) return true;
```

modules/server/src/main/ts/bedrock/cli/Hud.ts
```diff
@@ -25,24 +26,29 @@ export const create = (testfiles: string[], loglevel: 'simple' | 'advanced'): Hu
...
-    readline.clearLine(stream, 1);
+    if (!Env.IS_CI) {
+      readline.clearLine(stream, 1);
+    }
     return Promise.resolve();
   };
 
-  const advUpdate = (data: ResultData) => {
-    if (started) {
+  const advUpdate = async (data: ResultData) => {
```
ltrouton, line 41: I can't see any `await` or promises
```suggestion
  const advUpdate = (data: ResultData) => {
```

ltrouton (approved):

modules/server/src/main/ts/bedrock/server/Controller.ts
TheSpyder: What would be really nice is
* reduce frequency of client to HUD status push to 10% in CI
* when there is a test failure, override this and do a push immediately

TheSpyder: But this is at least _better_, so I think publish it and if you like work on that idea as a new PR.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Improved
+
+- Reduced progress logs when on CI to 10% intervals #TINY10708
```
TheSpyder, line 11:
```suggestion
- Reduced progress logs when on CI to 10% intervals #TINY-10708
```

PR #145, opened 2024-08-16
TINY-11177: Vastly improve remote testing

TheSpyder:
Related Ticket: TINY-11177

Description of Changes:
* Remote testing was slow because bedrock sent session POST requests to `/start` before each `it` block, and then again to `/results` after each `it` block. It would wait for these requests to finish before continuing.
* `/start` is now sent at startup only.
* `/results` is now only sent for failure and skip, or every 30 seconds if tests are passing happily.
* The bedrock UI doesn't block on these status requests anymore, it just throws them into an array and uses `Promise.all()` at the very end to wait for them.
* Unfortunately the `bedrock` CLI console HUD depended on receiving these status updates, so I had to adjust it to account for the missing reports.
* The bedrock HTML page now has a fake icon to remove the `favicon.ico` request which probably also slowed things down.
* A proper `keep-alive` header is now set, instead of using the default 5 seconds which was causing occasional `502 Bad Gateway` errors.
* Reverted #142 which was designed to solve this problem but server-side was the wrong answer.

Pre-checks:
* [x] Changelog entry added

Before merging:


spocke (approved):
How does this affect junit reporting would the in flight requests make them in random order or would things be skipped.

TheSpyder:
I posted in slack that this destroys JUnit reporting. Ideas for how to fix:
https://cksource.slack.com/archives/C04KXE6QUTX/p1724022794632699?thread_ts=1723766938.410059&cid=C04KXE6QUTX

Jenkinsfile
```diff
@@ -21,19 +21,35 @@ timestamps {
...
+        [ browser: 'chrome', provider: 'lambdatest', os: 'macOS Sonoma', buckets: 1 ],
+        [ browser: 'firefox', provider: 'lambdatest', os: 'macOS Sonoma', buckets: 1 ],
+        [ browser: 'safari', provider: 'lambdatest', os: 'macOS Sonoma', buckets: 1 ],
+      ],
+      prepareTests: {
+        yarnInstall()
+        sh 'yarn build'
+      },
+      testDirs: [ 'modules/sample/src/test/ts/**/pass' ],
+      custom: '--config modules/sample/tsconfig.json --customRoutes modules/sample/routes.json --polyfills Promise Symbol'
```
ltrouton, line 43: Can we remove the polyfills now since we are running on browsers that support Promise and Symbol?

modules/runner/src/main/ts/reporter/Reporter.ts
```diff
@@ -139,19 +179,51 @@ export const Reporter = (params: UrlParams, callbacks: Callbacks, ui: ReporterUi
...
   };
 
+  const waitForResults = async (): Promise<void> => {
+    sendCurrentResults();
+    if (requestsInFlight.length > 0) {
+      const currentRequests = requestsInFlight.slice(0);
+      requestsInFlight.length = 0;
+      return Promise.all(currentRequests).then(() => {
+        // if more things have been queued, such as a failing test stack trace, wait for those as well
+        waitForResults();
```
ltrouton, line 189: Should we be returning the promise from `waitForResults`? Although, I think since the function has the `async` identifier, JavaScript must handle it automagically.

modules/runner/src/main/ts/reporter/Reporter.ts
```diff
@@ -139,19 +179,54 @@ export const Reporter = (params: UrlParams, callbacks: Callbacks, ui: ReporterUi
     };
   };
 
+  const waitForResults = async (): Promise<void> => {
+    sendCurrentResults();
+    if (requestsInFlight.length > 0) {
```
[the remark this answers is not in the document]
TheSpyder: No, because it's an `async` function. That's done automatically.

Jenkinsfile
TheSpyder: I thought I tried that and it had some issues. But you're right it makes sense to wipe that out now!

TheSpyder: Turns out none of our editor automated tests use this anymore (a few manual still did). But it's still worth removing just in case!

TheSpyder:
For the record, JUnit records are long since fixed.

ltrouton (approved):
Not an expert on bedrock so don't know everything I am looking at but this looks like it will be a big improvement 👍

modules/runner/src/main/ts/reporter/Reporter.ts
```diff
@@ -1,22 +1,23 @@
-import { LoggedError, Reporter as ErrorReporter } from '@ephox/bedrock-common';
-import Promise from '@ephox/wrap-promise-polyfill';
-import { Callbacks } from './Callbacks';
-import { UrlParams } from '../core/UrlParams';
-import { formatElapsedTime, mapStackTrace, setStack } from '../core/Utils';
+import {LoggedError, Reporter as ErrorReporter} from '@ephox/bedrock-common';
```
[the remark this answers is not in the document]
TheSpyder: We don't have eslint running on this repository 😂 this is somehow the default in webstorm and I haven't worked out how to change it

modules/runner/src/main/ts/reporter/Reporter.ts
```diff
@@ -42,16 +43,41 @@ const mapError = (e: LoggedError) => mapStackTrace(e.stack).then((mappedStack) =
...
+      requestsInFlight.push(callbacks.sendTestResults(params.session, testResults));
+      testResults.length = 0;
+      return true;
+    }
+    return false;
+  };
+
+  const reportResult = (result: TestReport): void => {
+    testResults.push(result);
+    if (Date.now() - timeOfLastReport > 30 * 1000) {
```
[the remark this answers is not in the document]
TheSpyder: There are already configurable timeouts related to how long each test takes and how long the overall test suite can run.

This 30 second check relates to remote service timeouts; we're specifically avoiding calls to the remote server for performance reasons but that looks a lot like a hung test without regular status updates.

Bedrock has an internal (not configurable) timeout related to this, separate to the configurable test timeout, and LambdaTest also has their own. I haven't checked AWS but I expect it does as well.

modules/runner/src/main/ts/runner/TestRun.ts
```diff
@@ -92,6 +106,23 @@ export const runTests = (tests: Test[], state: RunState, actions: RunActions, re
   return loop(tests, (test) => runTest(test, state, actions, reporter));
 };
 
+const fakeFailure = (reporter: RunReporter, suite: Suite, state: RunState, siblings: string[], actions: RunActions) => {
+  const tests = suite.tests.length === 0 ? suite.suites[0].tests : suite.tests;
+  const filename = tests[0]?.file || 'Unknown';
```
spocke, line 111: Could we use a ternary here we try to avoid the `&&` and `||` hacks. I guess we have this style in other places in the project but should we still try to follow what we do in regular editor development?

modules/runner/src/main/ts/runner/TestRun.ts
```diff
@@ -52,17 +56,23 @@ export const runTest = (test: Test, state: RunState, actions: RunActions, report
...
-  const skip = (report: TestReporter) => {
+  const skip = (testReport: TestReporter) => {
     test.setResult(RunnableState.Skipped);
-    return report.skip(test.title).then(actions.onSkip);
+    testReport.skip(test.title);
+    actions.onSkip();
   };
 
-  const pass = (report: TestReporter) => {
+  const pass = async (testReport: TestReporter) => {
```
spocke, line 72: Why is this async when nothing inside it is async?

modules/runner/src/main/ts/runner/TestRun.ts
```diff
@@ -92,6 +106,23 @@ export const runTests = (tests: Test[], state: RunState, actions: RunActions, re
...
+  const fakeFailure = LoggedError.loggedError(new Error(`File ${filename} did not clean up after itself, extra elements were left in the DOM`), siblings);
+
+  // This is mostly duplicate of `fakeReporter.fail()`, but the "test" doesn't really exist
+  // so the UI gets confused and shows two errors if we do that
+  console.error(fakeFailure);
+  fakeReporter.fail(fakeFailure);
+  // this is where the page reloads if the global retry system is active
+  actions.onFailure();
+  // Test failures must be an empty reject, otherwise the error management thinks it's a bedrock error
+  return Promise.reject();
```
spocke, line 123: Hmm, I think you should use subclasses or Error if you want to distinct on error type here if possible since this seems fragile.

modules/runner/src/main/ts/runner/TestRun.ts
```diff
@@ -101,14 +132,33 @@ export const runSuite = (suite: Suite, state: RunState, actions: RunActions, rep
...
   }
 };
 
 export const runSuites = (suites: Suite[], state: RunState, actions: RunActions, reporter: RunReporter): Promise<void> => {
-  return loop(suites, (suite) => runSuite(suite, state, actions, reporter));
-};
\ No newline at end of file
+  return loop(suites, (suite) => {
+
+    const checkSiblings = !suite.parent?.root ? undefined :
```
spocke, line 155: This looks very strange and is hard to read use a if statement then branch with and without the then(checkSiblings). What does then(undefined) do I guess it's a noop.

modules/runner/src/test/ts/reporter/ReporterTest.ts
```diff
@@ -1,15 +1,15 @@
 import { Failure, LoggedError } from '@ephox/bedrock-common';
-import Promise from '@ephox/wrap-promise-polyfill';
 import { assert } from 'chai';
 import * as fc from 'fast-check';
 import { beforeEach, describe, it } from 'mocha';
 import { UrlParams } from '../../../main/ts/core/UrlParams';
 import { Callbacks, TestErrorData } from '../../../main/ts/reporter/Callbacks';
 import { Reporter } from '../../../main/ts/reporter/Reporter';
-import { noop } from '../TestUtils';
+import {noop, wait} from '../TestUtils';
```
spocke, line 8: Weird formatting again. WebStorm thing?

modules/server/src/main/ts/bedrock/auto/Driver.ts
```diff
@@ -1,15 +1,14 @@
-
 import * as path from 'path';
 import * as childProcess from 'child_process';
 import * as os from 'os';
 import * as WebdriverIO from 'webdriverio';
+import {RemoteOptions} from 'webdriverio';
```
spocke, line 5: Formatting.

modules/server/src/main/ts/bedrock/auto/RemoteDriver.ts
```diff
@@ -4,6 +4,8 @@ import * as deepmerge from 'deepmerge';
 import { DeviceFarmClient, CreateTestGridUrlCommand } from '@aws-sdk/client-device-farm';
 import { Driver, DriverSettings } from './Driver';
 
+export const REMOTE_IDLE_TIMEOUT_SECONDS = 360;
```
spocke, line 7: I wonder if this is something that we might want to configure?

modules/server/src/main/ts/bedrock/server/Apis.ts
```diff
@@ -119,7 +122,26 @@ export const create = (master: DriverMaster | null, maybeDriver: Attempt<any, Br
...
+
+  const maybeSendKeepAlive: () => Promise<void> = (() => {
+    let lastKeepAlive = Date.now();
+    /*
+     * If we're within a minute of the remote driver timeout, send a keep-alive.
+     *
+     * In theory this is only required if no other action has been taken in the last x minutes,
+     * but this happens so rarely it should be fine.
+     */
+    const keepAliveTimer = (Math.max(120, REMOTE_IDLE_TIMEOUT_SECONDS - 60)) * 1000;
```
spocke, line 135:
```suggestion
    const keepAliveTimer = Math.max(120, REMOTE_IDLE_TIMEOUT_SECONDS - 60) * 1000;
```

modules/server/src/main/ts/bedrock/server/Controller.ts
```diff
@@ -214,20 +218,16 @@ export const create = (stickyFirstSession: boolean, singleTimeout: number, overa
...
             } else if (allElapsed > overallTimeout) {
               // combined tests took too long
               let lastTest;
-              if (session.inflight !== null) {
-                const runningTime = now - session.inflight.start;
-                lastTest = 'Current test: ' + testName(session.inflight) + ' running ' + formatTime(runningTime) + '.';
-              } else if (session.previous !== null) {
+              if (session.previous !== null) {
                 const sincePrevious = now - session.previous.end;
                 lastTest = 'Previous test: ' + testName(session.previous) + ' finished ' + formatTime(sincePrevious) + ' ago.';
```
spocke, line 232: Could we some how dump the current test here as well? Sometimes I noticed test being stuck and timing out then it hard to figure out what caused it.

modules/server/src/main/ts/bedrock/server/EffectUtils.ts
```diff
@@ -39,8 +39,10 @@ const performActionOnFrame = async <T>(driver: Browser, selector: string, action
     await driver.switchToFrame(null);
     return result;
   } catch (err: any) {
-    await driver.switchToFrame(null);
-    return Promise.reject(err);
+    return driver.status().then(status => {
+      console.log('webdriver failed with status', status);
```
spocke, line 43: Should this be a console.error?

modules/server/tasks/bedrock.js
```diff
@@ -68,7 +69,8 @@ module.exports = function(grunt) {
     autoSettings.gruntDone = function (passed) {
       done(passed);
     };
-    autoSettings.stopOnFailure = options.stopOnFailure;
+    autoSettings.stopOnFailure = options.stopOnFailure || autoSettings.stopOnFailure;
```
spocke, line 72: There is a Type module in server that can be used for ternaries on these things if we want to avoid `||` fallbacks.

spocke (approved):
Looks good, lots of nice improvements beyond the perf things. 👍

modules/runner/src/test/ts/reporter/ReporterTest.ts
TheSpyder: yep.

modules/server/src/main/ts/bedrock/auto/Driver.ts
TheSpyder: One day we'll enable eslint on the project 😂

modules/server/src/main/ts/bedrock/auto/RemoteDriver.ts
TheSpyder: I think that would make configuring bedrock too complex.

LambdaTest defaults to 100 seconds, and this is hard coded (then referenced elsewhere) to make sure we don't time out the LambdaTest run while waiting for our configurable timeouts.

modules/server/src/main/ts/bedrock/server/Apis.ts
TheSpyder: WebStorm used to have a setting to identify optimisations like this. I wonder why it's not on by default.

modules/server/src/main/ts/bedrock/server/Controller.ts
TheSpyder: This is a whole test suite timeout, not a single test stuck timeout. The single test stuck timeout is now completely client side.

The server cannot really rely on the contents of `session` here since it's updated so infrequently now. I kept printing the previous test just so we at least have _something_ that might be useful.

modules/server/src/main/ts/bedrock/server/EffectUtils.ts
TheSpyder: Probably. I don't think I hit this very much in my testing.

TheSpyder:
Resolved all except one which I'll deal with on Monday and then release this 👍

modules/server/tasks/bedrock.js
spocke: Right that would be harder. It's a bummer we can't use katamari on these things.

TheSpyder:
I'm glad to finally finish this one ☺️

PR #168, opened 2026-03-24
TINY-14141: Add `skipTypecheck` CLI option

shanmen-tiny:
Related Ticket: TINY-14141

Description of Changes:
- Add a `--skipTypecheck` CLI option to disable typechecking before tests when it is already being run separately, such as in CI

Pre-checks:
* [x] Changelog entry added

Before merging:


modules/server/src/test/ts/ClisTest.ts
```diff
@@ -57,6 +57,7 @@ describe('Clis.forAuto', () => {
       verbose: false,
       bucket: 1,
       buckets: 1,
+      skipTypecheck: false,
```
kimwoodfield, line 60: Do we need to provide these if it's `false` by default? 🤔

kimwoodfield (approved):

TheSpyder (approved):
When you're integrating this remember that quite often we use bedrock through grunt instead of command line options directly.

Bedrock's grunt task will include all exported `ClOptions` so this is fine, but it's worth keeping in mind.

modules/server/src/main/ts/bedrock/server/RunnerRoutes.ts
```diff
@@ -18,8 +18,43 @@ interface WorkspaceRoot {
   folder: string;
 }
 
-export const generate = async (mode: string, projectdir: string, basedir: string, scratchdir: string, configFile: string, bundler: Types.Bundler, testfiles: string[], chunk: number,
```
spocke, line 21: Nice cleanup, a good example when people just add and add their own duct tape to the ball.

spocke (approved):

modules/server/src/test/ts/ClisTest.ts
shanmen-tiny: That's required, it's asserting the result 👍

